PR #5, opened 2022-04-19
springer: Migrate files from SFTP to S3

Neodia:
* Added S3 service in docker-compose for local dev

* Added common services that will be used accross all publishers

* Added DAG for Springer to get the files from SFTP and push them to S3

* Added tests

* ref: #2


dags/common/pull_ftp.py
```diff
@@ -0,0 +1,25 @@
+
```
drjova: could you please install `pre-commit` https://pre-commit.com we have already a configuration for code styling

dags/common/pull_ftp.py
```diff
@@ -0,0 +1,25 @@
...
+
+import io
+import zipfile
+from common.repository import IRepository
+from common.sftp_service import SFTPService
+
+
+def migrate_from_ftp(sftp: SFTPService, repo: IRepository):
```
drjova, line 8: I would remove completely the types, since we don't use them in our projects and we don't have any tool to check them (i.e. `mypy`)

dags/springer/sftp_service.py
```diff
@@ -0,0 +1,18 @@
+import pathlib
+import yaml
+
+from common.sftp_service import SFTPService
+
+
+class SpringerSFTPService(SFTPService):
+    def __init__(self):
+        config_path : pathlib.Path = pathlib.Path(__file__).parent.parent.resolve()
```
drjova: could we use env vars instead of config files?

requirements.txt
```diff
@@ -1 +1,10 @@
-pre-commit==2.17.0
+apache-airflow==2.2.4 --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.2.4/constraints-3.8.txt
```
drjova, line 2: why did you remove pre-commit? 🤔

tests/test_dag/test_dag.py
```diff
@@ -1,9 +1,24 @@
...
+
+
+def test_task1():
+    value = task_1()
+    assert value == "first_task"
+
+
+def test_task2():
+    value = task_2()
+    assert value == "second_task"
```
drjova: we don't need those any more, you can remove them with the test_dag file as well

dags/springer/repository.py
```diff
@@ -0,0 +1,54 @@
...
+        with open(f"{config_path}/config.yaml", "r") as config_file:
+            config = yaml.safe_load(config_file)['publishers']['springer']['repo']
+            self.s3 = boto3.resource(
+                's3',
+                endpoint_url=config['source'],
+                aws_access_key_id=config['username'],
+                aws_secret_access_key=config['password'],
+                config=Config(signature_version='s3v4'),
+                region_name=config['region']
+            ).Bucket(config['bucket'])
```
drjova: I would abstract also the use of boto configuration

dags/springer/repository.py
```diff
@@ -0,0 +1,54 @@
...
+            last_part = file_parts[-1]
+            filename_without_extension = last_part.split('.')[0]
+            if not filename_without_extension in ret_dict.keys():
+                ret_dict[filename_without_extension] = dict()
+            ret_dict[filename_without_extension]['xml' if self.__file_is_meta(last_part) else 'pdf'] = file
+        return list(ret_dict.values())
+
+    def save(self, filename: str, obj: IO):
+        prefix = self.ZIPED_DIR if '.zip' in filename else self.EXTRACTED_DIR
+        self.s3.upload_fileobj(obj, prefix + filename)
```
drjova, line 38: what is the final structure on the s3 bucket for springer?

drjova (commented):
Thanks! Please have a look at the comments. Also in python 3 you don't need `__init__.py` files.

dags/common/pull_ftp.py
Neodia: I think the types are a great tool to have to indicate to the dev what we expect as an input. Furthermore, I find them useful when developing, as I get IntelliSense on the variable that allows me to code without going back and forth. What do you think ? I come from strongly typed languages so I might be biased

dags/springer/sftp_service.py
Neodia: Yes for sure. I didn't know how the config was injected in the team so I went with this.

requirements.txt
Neodia: My bad, adding it back following your previous comment

dags/springer/repository.py
Neodia: Abstracting the boto config will lock us later if we want to switch the repository source. Could it happen later ? If not I'll put the boto config in the interface aswell 👍

dags/springer/repository.py
Neodia: Screenshot 2022-04-20 at 14 05 33
Does that seem appropriated ?

dags/common/pull_ftp.py
Neodia: Done

requirements.txt
Neodia: Done

tests/test_dag/test_dag.py
Neodia: Done

dags/springer/sftp_service.py
Neodia: Done

dags/common/pull_ftp.py
drjova: Indeed the types are a good in many cases, although we don't use them in the team, I would recommend not to introduce them at this point on python our applications.

dags/springer/repository.py
drjova: good point, I don't see changing it in the long term, we recently switched to s3 and it's used across the applications

dags/springer/repository.py
drjova: 👍 my concern is that we will not be able to track which zip files correspond to extracted files, I propose to use `extracted/ftp_PUB_19-01-XXXXXX/JOU=XXXX` etc in order to be able to easily traceback the packages.

Neodia: Screenshot 2022-04-20 at 16 29 42
Better ?

.github/workflows/test-and-build.yml
```diff
@@ -16,13 +16,17 @@ jobs:
...
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           pip install -r requirements-airflow.txt -r requirements.txt -r requirements-test.txt
-      - name: Re-create unit test sqlite database
-        run: airflow db reset -y
+      - name: Run services for IT tests
+        run: make start
+      - name: Wait for docker services
+        run: sleep 10
```
drjova: do we really need the sleep?

dags/springer/repository.py
```diff
@@ -0,0 +1,55 @@
...
+from typing import IO
+
+import boto3
+from botocore.client import Config
+from common.repository import IRepository
+from mypy_boto3_s3.service_resource import _Bucket
+
+
+class SpringerRepository(IRepository):
+    ZIPED_DIR: str = "raw/"
```
drjova, line 10: there are still some types left 😅

dags/springer/dag_pull_ftp.py
```diff
@@ -0,0 +1,26 @@
+import airflow
+import common.pull_ftp as pull_ftp
```
drjova, line 2: Why not just import directly the `trigger_file_processing` function?

dags/springer/repository.py
```diff
@@ -0,0 +1,55 @@
...
+    def __init__(self) -> None:
+        super().__init__()
+        self.s3 = boto3.resource(
+            "s3",
+            endpoint_url=os.getenv("SPRINGER_REPO_SOURCE", "http://localhost:9000"),
+            aws_access_key_id=os.getenv("SPRINGER_REPO_USERNAME", "airflow"),
+            aws_secret_access_key=os.getenv("SPRINGER_REPO_PASSWORD", "Airflow01"),
+            config=Config(signature_version="s3v4"),
+            region_name=os.getenv("SPRINGER_REPO_REGION", "us-east-1"),
+        ).Bucket(os.getenv("SPRINGER_REPO_BUCKET", "springer"))
```
drjova: Sorry if I wasn't clear, I meant to have a service for s3 that we initialize with the same config and just configure the bucket per publisher.

integration/common/test_IT_sftp.py
```diff
@@ -0,0 +1,24 @@
+import io
```
drjova: since we have  a separate `integration` directory we don't need the `IT` prefix 😄

drjova (commented):
@Neodia thanks for the update, it looks good well done. Few more things and it's ready to 🚀 :
* could you please squash your commits?
* could you please remove the types 😃 
* tests structure should be `tests/unit` `tests/integration`

.github/workflows/test-and-build.yml
Neodia: It's really not ideal but I was having trouble in the pipeline when not adding them

dags/springer/dag_pull_ftp.py
Neodia: Because the internal function has the same name. Thus, it will to an indefinite recursive call instead.

dags/springer/repository.py
Neodia: Ohh got you, will do

dags/springer/repository.py
Neodia: As discussed, we will discuss types as a team. Until then we keep those type hints 👍

integration/common/test_IT_sftp.py
Neodia: Done

.github/workflows/test-and-build.yml
drjova: We can use gh services instead

dags/springer/repository.py
Neodia: Done

dags/common/s3_service.py
```diff
@@ -0,0 +1,15 @@
+import os
+
+import boto3
+from botocore.client import Config
+
+
+def S3Service(bucket: str):
+    return boto3.resource(
+        "s3",
+        endpoint_url=os.getenv("REPO_SOURCE", "http://localhost:9000"),
```
drjova:
```suggestion
        endpoint_url=os.getenv("S3_ENDPOINT", "http://localhost:9000"),
```

dags/common/s3_service.py
```diff
@@ -0,0 +1,15 @@
...
+        aws_access_key_id=os.getenv("REPO_USERNAME", "airflow"),
```
drjova:
```suggestion
        aws_access_key_id=os.getenv("S3_USERNAME", "airflow"),
```

dags/common/s3_service.py
```diff
@@ -0,0 +1,15 @@
...
+        aws_secret_access_key=os.getenv("REPO_PASSWORD", "Airflow01"),
```
drjova:
```suggestion
        aws_secret_access_key=os.getenv("S3_PASSWORD", "Airflow01"),
```

dags/common/s3_service.py
```diff
@@ -0,0 +1,15 @@
...
+        config=Config(signature_version="s3v4"),
+        region_name=os.getenv("REPO_REGION", "us-east-1"),
```
drjova:
```suggestion
        region_name=os.getenv("S3_REGION", "us-east-1"),
```

drjova (commented):
LGTM just the comment for the s3 env vars and it's ready to 🚀

drjova (approved):
🥇

.github/workflows/test-and-build.yml
Neodia: As discussed, the GitHub actions Services does not have custom commands for now. Thus, keeping the sleep command for the pipeline

PR #6, opened 2022-04-21
6 aps migration

ErnestaP:
Screen Shot 2022-04-29 at 16 58 06

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ export AIRFLOW_HOME=/path/to/cloned/repo
 ```
 
 4. Install all required dependencies for a project. We need 3 requirements files, when we're running the project locally. First file (requirements.py) has all additional dependencies required to run the tasks correctly (connecting to ftputil, boto3 and etc.), the second one installs requirememnts for testing, such as pytest, the third- (requirements_airflow.py) installs Airflow itself and constraints needed for it. The the third file is not needed when we're running the procject on Docker, because we're using Apache Airflow Docker image:
-   `pip install requirements.txt -r requirements-test.txt -r requirements-airflow.txt`
+   `pip install -r requirements.txt -r requirements-test.txt -r requirements-airflow.txt`
 5. Run Airflow. Airflow comes with the `standalone` command. This should not be used as it seems to use the SequentialExecutor by default. To have our local config being use, we must run all services together :
 
 ```


dags/aps/APSParams.py
```diff
@@ -0,0 +1,23 @@
...
+        self.from_date = from_date
+        self.until_date = until_date
+        self.date = date
+        self.journals = journals
+        self.sets = sets
+        self.per_page = per_page
+
+    def get_params(self) -> dict:
+        params = {'from': self.from_date, 'until': self.until_date, 'date': self.date, 'journals': self.journals, 'sets': self.sets, 'per_page': self.per_page}
+        print(params)
```
drjova: is tis for debugging?

dags/aps/constants.py
```diff
@@ -0,0 +1,4 @@
+END_POINT_FOR_S3='https://s3.cern.ch'
+BASE_URL_FOR_APS='http://harvest.aps.org/v2/journals/articles'
+ARTICLE_BASE_URL_FOR_APS='http://harvest.aps.org/v2/journals/articles'
+BUCKET_NAME='scoap3-test-ernesta'
```
drjova: These should be environment vars

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,57 @@
...
+    today = date.today().strftime("%Y-%m-%d_%H:%M:%S")
+    prefix = os.path.join(root, today)
+    json_obj = json.loads(response.content)
+
+    if response.status_code == 200 and len(json_obj['data']) > 0:
+        key = '.'.join([prefix, 'json'])
+        byte_file = io.BytesIO(response.content)
+        repo.save(key, byte_file)
+        return key
+    return None
```
drjova: by default the function returns `None` no need to add `return`

dags/aps/repository.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import IO
+from datetime import date
+
+from common.repository import IRepository
+from common.s3_service import S3Service
+from dotenv import load_dotenv
+from datetime import timedelta
+
+## Loading virtual var.
+load_dotenv()
```
drjova: why `dotenv`? We don't need it since we export the env vars

dags/common/pull_ftp.py
```diff
@@ -9,13 +9,13 @@ def migrate_from_ftp(sftp: SFTPService, repo: IRepository):
...
-                repo.save(file_prefix + "/" + zip_filename, io.BytesIO(file_content))
-        repo.save(file, file_bytes)
-
+        if zipfile.is_zipfile(file_bytes):
+            with zipfile.ZipFile(file_bytes) as zip:
+                    for zip_filename in zip.namelist():
+                        file_prefix = file.split(".")[0]
+                        file_content = zip.read(zip_filename)
+                        repo.save(file_prefix + "/" + zip_filename, io.BytesIO(file_content))
+            repo.save(file, file_bytes)
```
drjova:
```suggestion
        if not zipfile.is_zipfile(file_bytes):
            continue
        with zipfile.ZipFile(file_bytes) as zip:
                for zip_filename in zip.namelist():
                    file_prefix = file.split(".")[0]
                    file_content = zip.read(zip_filename)
                    repo.save(file_prefix + "/" + zip_filename, io.BytesIO(file_content))
        repo.save(file, file_bytes)
```

dags/common/rest_api_service.py
```diff
@@ -0,0 +1,35 @@
...
+        
+    def get_response(self):
+        from furl import furl
+        import requests
+
+        url = furl(self.base_url).add(self.parameters).url
+        if bool(self.headers):
+            return requests.get(url, headers=self.headers)
+        else:
+            return requests.get(url)
```
drjova:
```suggestion
        url = furl(self.base_url).add(self.parameters).url
        if bool(self.headers):
            return requests.get(url, headers=self.headers)
        return requests.get(url)
```

dags/common/rest_api_service.py
```diff
@@ -0,0 +1,35 @@
...
+        self.base_url = base_url 
+        self.parameters = parameters
+        self.headers = headers
+
+    def __repr__(self) -> str:
+        return(f"< RestApi base_url={self.base_url}, parameters={self.parameters}, header={self.headers} >")
+        
+    def get_response(self):
+        from furl import furl
+        import requests
```
drjova: why importing inside the function?

Makefile
```diff
@@ -43,4 +43,10 @@ stop:
 	cat $(WORKER_PID) | xargs kill
 	cat $(FLOWER_PID) | xargs kill
 	rm *.out *.err *.log
+	kill -9 $(lsof -ti:8080)
+	rm *.out *.err
 	echo -e "\033[0;32m Airflow Stoped. \033[0m"
+
+clean:
+	docker-compose down --volumes --remove-orphans
+	docker rmi $(docker images -a -q) --force
```
drjova: this will delete all your local images, why we would want this?

dags/aps/APSParams.py
ErnestaP: yes

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def set_APS_harvesting_interval(repo):
+    today = date.today()
+    aps_fetching_from_date = repo.find_the_last_uploaded_file_date('APS')
+    aps_fetching_until_date = today.strftime("%Y-%m-%d")
+    if aps_fetching_from_date is None:
+        Variable.set('aps_fetching_from_date', aps_fetching_until_date)
+    else:
+        Variable.set('aps_fetching_from_date', aps_fetching_from_date)
```
Neodia:
```suggestion
    Variable.set('aps_fetching_from_date', aps_fetching_until_date if aps_fetching_until_date is None else aps_fetching_from_date)
```

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,57 @@
...
+    Variable.set('aps_fetching_until_date', aps_fetching_until_date)
+    return({'aps_fetching_from_date': Variable.get('aps_fetching_from_date'), 'aps_fetching_until_date': Variable.get('aps_fetching_until_date')})
```
Neodia: I think this can be less verbose by using variables to store this data, and only using Variable.get API to get the data when we do not have it

dags/aps/APSParams.py
drjova: you could use logging for that with `debug` level

dags/aps/repository.py
```diff
@@ -0,0 +1,44 @@
...
+
+class APSRepository(IRepository):
+    def __init__(self) -> None:
+        super().__init__()
+        self.s3_bucket = S3Service(os.getenv('BUCKET_NAME_APS', 'aps'))
+
+    def find_all(self):
+        files = []
+        for obj in self.s3_bucket.objects.filter(Prefix='APS').all():
```
Neodia: Why is this filter needed if we already are inside the APS bucket ?

dags/common/rest_api_service.py
ErnestaP: https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#top-level-python-code

drjova: good to know thanks 👍

Makefile
ErnestaP: Just to start everything from scratch, to have DAGs without runs

dags/common/rest_api_service.py
```diff
@@ -0,0 +1,35 @@
+class RestApi:
```
Neodia: I think we should review how this class is being used.

It's a detail but this class is not a RestApi but a RestApiClient, the API il already being served elsewhere.

I think the class does a good job in being generic, but it should not be used as it currently is. Here is why :
- What you are handling are "responses". I think it's better to think in business logic when you can. Our repositories are not handling responses, they're handling files.

I think this class should be used a a base (maybe with a has-a relationship instead of is-a) for the other API clients. We should have a API client for APS and one for each publisher that needs it. Each API client should know how to configure itself, and should serve functions that allows us to directly understand what we are getting from the client.

The pieces of code that will use the API client don't care about headers and parameters, they care about "give me the articles jsons from last month to today", and I think that's how we should design this service

I hope I was clear enough 😆

dags/aps/aps_api_client.py
```diff
@@ -0,0 +1,43 @@
...
+import json
+
+from common.rest_api_client import RestApiClient
+
+
+class APSApiClient(RestApiClient):
+    def __init__(self):
+        super().__init__()
+        self.base_url = getenv("BASE_URL_FOR_APS", 'http://harvest.aps.org/v2/journals/articles')
+        self.rest_api_client = RestApiClient(base_url=self.base_url)
```
Neodia: There is some weird relationship between those classes here. APSApiClient `is` a RestApiClient (as it is the base class) and it also `has` a RestApiClient

dags/aps/aps_api_client.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def get_articles_metadata(self):
+        response = self.rest_api_client.get_response()
+        if response.status_code == 200:
```
Neodia: Nothing happens if the status_code isn't `ok`. We should add output

dags/aps/aps_api_client.py
```diff
@@ -0,0 +1,43 @@
...
+            if len(json_obj['data']) > 0:
+                return response.content
+            return ''
+
+    def get_pdf_file(self, doi):
+        headers = {'Accept': 'application/pdf'}
+        self.rest_api_client.subdir = doi
+        response = self.rest_api_client.get_response()
+        self.rest_api_client.set(base_url=self.base_url, headers=headers)
+        if response.status_code == 200:
```
Neodia: Same here

dags/aps/aps_api_client.py
```diff
@@ -0,0 +1,43 @@
...
+            return response.content
+
+    def get_xml_file(self, doi):
+        headers = {'Accept': 'text/xml'}
+        self.rest_api_client.subdir = doi
+        response = self.rest_api_client.get_response()
+        self.rest_api_client.set(base_url=self.base_url, headers=headers)
+        if response.status_code == 200:
```
Neodia: Same here. You are using the RestApiClient so maybe you could wrap the error handling there ?

dags/aps/aps_api_client.py
```diff
@@ -0,0 +1,43 @@
...
+            return response.content
+
+    def get_response(self):
```
Neodia: Why is this needed ?

dags/aps/aps_api_client.py
```diff
@@ -0,0 +1,43 @@
...
+        self.get_response(self)
+
+    def set(self, base_url="", parameters={}, headers={}):
```
Neodia: Why do we have to set stuff here ?

dags/common/rest_api_client.py
```diff
@@ -1,13 +1,15 @@
-class RestApi:
+class RestApiClient:
```
Neodia: Now that I see it, this class is called `RestApiClient` but the usage that you have inside it is the one of a `Request`. That's why you are setting parameters and headers, you are treating this class as a `Request`. I think this is fine but you must chose a path between both. I think for our use case, It's better to go with the Request. So we have : 
- common/request
  - Here we have an object that gets the base path, port, request parameters and headers
  - We can then run this request. The error handling should be done in this class, and the result should be returned
- aps/APSApiClient
  - This client uses the request class when triggering requests. The error handling is in the request object, so the class can focus on only getting the data when it exists and doing what it must with it

PR #9, opened 2022-04-27
Springer: Added Parsing

Neodia:
@drjova @ErnestaP

### DIFF
--- a/dags/common/parsing/parser.py
+++ b/dags/common/parsing/parser.py
@@ -1,0 +1,29 @@
+import xml.etree.ElementTree as ET
+from typing import List
+
+from common.parsing.extractor import IExtractor
+
+
+class IParser:
+    extractors: List[IExtractor]
+
+    def __init__(self, extractors) -> None:
+        self.extractors = extractors
+
+    def parse(self, article: ET.Element):
+        return {
+            extractor.destination: value
+            for extractor in self.extractors
+            if (value := extractor.extract(article)) is not None
+        }
+
+
+class ObjectExtractor(IParser, IExtractor):
+    def __init__(self, destination, extractors, extra_function=lambda x: x) -> None:
+        super().__init__(destination)
+        self.destination = destination
+        self.extractors = extractors
+        self.extra_function = extra_function
+
+    def extract(self, article: ET.Element):
+        return self.extra_function(super().parse(article))
--- a/dags/springer/dag_process_file.py
+++ b/dags/springer/dag_process_file.py
@@ -1,0 +1,31 @@
+import base64
+import xml.etree.ElementTree as ET
+
+import airflow
+from airflow.decorators import dag, task
+from springer.parser import SpringerParser
+
+
+def springer_parse_file(**kwargs):
+    if "params" in kwargs and "file" in kwargs["params"]:
+        encoded_xml = kwargs["params"]["file"]
+        xml_bytes = base64.b64decode(encoded_xml)
+        xml = ET.fromstring(xml_bytes.decode("utf-8"))
+
+        parser = SpringerParser()
+        parsed = parser.parse(xml)
+
+        return parsed
+    raise Exception("There was no 'file' parameter. Exiting run.")
+
+
+@dag(start_date=airflow.utils.dates.days_ago(0))
+def springer_process_file():
+    @task()
+    def parse_file(**kwargs):
+        return springer_parse_file(**kwargs)
+
+    parse_file()
+
+
+dag_taskflow = springer_process_file()
--- a/dags/springer/parser.py
+++ b/dags/springer/parser.py
@@ -1,0 +1,257 @@
+import datetime
+import re
+import xml.etree.ElementTree as ET
+
+from common.parsing.parser import IParser, ObjectExtractor
+from common.parsing.xml_extractors import (
+    AttributeExtractor,
+    CustomExtractor,
+    TextExtractor,
+)
+
+
+class SpringerParser(IParser):
+    def __init__(self) -> None:
+
+        article_type_mapping = {
+            "OriginalPaper": "article",
+            "ReviewPaper": "review",
+            "BriefCommunication": "article",
+            "EditorialNotes": "editorial",
+            "BookReview": "review",
+            "ContinuingEducation": "other",
+            "Interview": "other",
+            "Letter": "other",
+            "Erratum": "erratum",
+            "Legacy": "other",
+            "Abstract": "other",
+            "Report": "other",
+            "Announcement": "other",
+            "News": "other",
+            "Events": "other",
+            "Acknowledgments": "other",
+            "MediaReport": "other",
+            "BibliographicalNote": "other",
+            "ProductNotes": "other",
+            "Unknown": "other",
+        }
+
+        extractors = [
+            AttributeExtractor(
+                "journal_doctype",
+                "./Journal/Volume/Issue/Article/ArticleInfo",
+                "ArticleType",
+                extra_function=lambda x: article_type_mapping[x],
+            ),
+            TextExtractor(
+                "dois", "./Journal/Volume/Issue/Article/ArticleInfo/ArticleDOI"
+            ),
+            CustomExtractor("arxiv_eprints", self._get_arxiv_eprints),
+            CustomExtractor("page_nr", self._get_page_nrs),
+            CustomExtractor("abstract", self._get_abstract),
+            TextExtractor(
+                "title",
+                "./Journal/Volume/Issue/Article/ArticleInfo/ArticleTitle",
+            ),
+            CustomExtractor("authors", self._get_authors),
+            TextExtractor(
+                "collaborations",
+                "./Journal/Volume/Issue/Article/ArticleHeader/AuthorGroup/InstitutionalAuthor/InstitutionalAuthorName",
+                False,
+                extra_function=lambda x: [x],
+            ),
+            TextExtractor(
+                "journal_title",
+                "./Journal/JournalInfo/JournalTitle",
+                extra_function=lambda s: s.lstrip("The "),
+            ),
+            TextExtractor(
+                "journal_issue", "./Journal/Volume/Issue/IssueInfo/IssueIDStart"
+            ),
+            TextExtractor(
+                "journal_volume", "./Journal/Volume/VolumeInfo/VolumeIDStart"
+            ),
+            AttributeExtractor("journal_artid", "./Journal/Volume/Issue/Article", "ID"),
+            TextExtractor(
+                "journal_fpage",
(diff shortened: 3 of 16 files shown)


dags/common/parsing/parser.py
```diff
@@ -0,0 +1,29 @@
...
+    extractors: List[IExtractor]
+
+    def __init__(self, extractors) -> None:
+        self.extractors = extractors
+
+    def parse(self, article: ET.Element):
+        return {
+            extractor.destination: value
+            for extractor in self.extractors
+            if (value := extractor.extract(article)) is not None
```
drjova, line 17: 💯

dags/springer/parser.py
```diff
@@ -0,0 +1,259 @@
...
+                    "./Journal/Volume/Issue/Article/ArticleHeader/AuthorGroup"
+                ),
+                contrib,
+            )
+
+            authors.append(author)
+
+        # if not authors:
+        #     logger.error('No authors found for article %s.' % dois)
+
```
drjova: leftover?

dags/springer/parser.py
```diff
@@ -0,0 +1,259 @@
...
+        return authors
+
+    def _get_page_nrs(self, article: ET.Element):
+        first_page = article.find(
+            "./Journal/Volume/Issue/Article/ArticleInfo/ArticleFirstPage"
+        ).text
+        last_page = article.find(
+            "./Journal/Volume/Issue/Article/ArticleInfo/ArticleLastPage"
+        ).text
+        return [int(last_page) - int(first_page) + 1]
```
drjova: what if last_page or first_page is empty string?

tests/integration/springer/test_dag_process_file.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def test_dag_loaded(dag: DAG):
+    assert dag is not None
+    assert len(dag.tasks) == 1
+
+
+def test_dag_run(dag: DAG, article):
+    id = datetime.datetime.utcnow().strftime(
+        "test_springer_dag_process_file_%Y-%m-%dT%H:%M:%S.%f"
```
drjova, line 48: I would use for all tests [freezegun](https://github.com/spulec/freezegun) to have always the same results and not depend on the current time.

tests/units/springer/test_parser.py
```diff
@@ -0,0 +1,346 @@
...
+    """Test extracting dois."""
+    dois = (
+        "10.1007/JHEP01(2019)210",
+        "10.1140/epjc/s10052-019-6572-3",
+        "10.1140/epjc/s10052-019-6540-y",
+    )
+    for doi, article in zip(dois, parsed_articles):
+        assert "dois" in article
+        assert article["dois"] == doi
+        break
```
drjova: why do you need break?

dags/springer/dag_process_file.py
```diff
@@ -0,0 +1,32 @@
...
+        encoded_xml = kwargs["params"]["file"]
+        xml_bytes = base64.b64decode(encoded_xml)
+        xml = ET.fromstring(xml_bytes.decode("utf-8"))
+
+        parser = SpringerParser()
+        parsed = parser.parse(xml)
+
+        return parsed
+    else:
+        raise Exception("There was no 'file' parameter. Exiting run.")
```
drjova:
```suggestion
def springer_parse_file(**kwargs):
    if "params" in kwargs and "file" in kwargs["params"]:
        encoded_xml = kwargs["params"]["file"]
        xml_bytes = base64.b64decode(encoded_xml)
        xml = ET.fromstring(xml_bytes.decode("utf-8"))

        parser = SpringerParser()
        parsed = parser.parse(xml)

        return parsed
    raise Exception("There was no 'file' parameter. Exiting run.")
```

dags/springer/parser.py
```diff
@@ -0,0 +1,259 @@
...
+            TextExtractor(
+                "collections",
+                "./Journal/JournalInfo/JournalTitle",
+                extra_function=lambda x: [x.lstrip("The ")],
+            ),
+        ]
+        super().__init__(extractors)
+
+    def _get_abstract(self, article: ET.Element):
+        paragraph = article.find(
```
drjova, line 112: what's the structure of `Para` could we just get the whole content instead of iterating multiple times?

Neodia:
```xml
                     <Para ID="Par1">
                        This paper describes a strategy for a general search used by the ATLAS Collaboration to find potential indications of new physics. Events are classified according to their final state into many event classes. For each event class an automated search algorithm tests whether the data are compatible with the Monte Carlo simulated expectation in several distributions sensitive to the effects of new physics. The significance of a deviation is quantified using pseudo-experiments. A data selection with a significant deviation defines a signal region for a dedicated follow-up analysis with an improved background expectation. The analysis of the data-derived signal regions on a new dataset allows a statistical interpretation without the large look-elsewhere effect. The sensitivity of the approach is discussed using Standard Model processes and benchmark signals of new physics. As an example, results are shown for 3.2 fb
                        <InlineEquation ID="IEq1">
                           <InlineMediaObject>
                              <ImageObject Color="BlackWhite" FileRef="10052_2019_6540_Article_IEq1.gif" Format="GIF" Rendition="HTML" Type="Linedraw" />
                           </InlineMediaObject>
                           <EquationSource Format="TEX">
                              <![CDATA[$$^{-1}$$]]>
                           </EquationSource>
                           <EquationSource Format="MATHML">
                              <![CDATA[
                    ]]>
                              <math xmlns:xlink="http://www.w3.org/1999/xlink">
                                 <msup xmlns:xlink="http://www.w3.org/1999/xlink">
                                    <mrow xmlns:xlink="http://www.w3.org/1999/xlink" />
                                    <mrow xmlns:xlink="http://www.w3.org/1999/xlink">
                                       <mo xmlns:xlink="http://www.w3.org/1999/xlink">-</mo>
                                       <mn xmlns:xlink="http://www.w3.org/1999/xlink">1</mn>
                                    </mrow>
                                 </msup>
                              </math>
                              <![CDATA[
                  ]]>
                           </EquationSource>
                        </InlineEquation>
                        of proton–proton collision data at a centre-of-mass energy of 13 
                        <InlineEquation ID="IEq2">
                           <InlineMediaObject>
                              <ImageObject Color="BlackWhite" FileRef="10052_2019_6540_Article_IEq2.gif" Format="GIF" Rendition="HTML" Type="Linedraw" />
                           </InlineMediaObject>
                           <EquationSource Format="TEX">
                              <![CDATA[$$\text {TeV}$$]]>
                           </EquationSource>
                           <EquationSource Format="MATHML">
                              <![CDATA[
                    ]]>
                              <math xmlns:xlink="http://www.w3.org/1999/xlink">
                                 <mtext xmlns:xlink="http://www.w3.org/1999/xlink">TeV</mtext>
                              </math>
                              <![CDATA[
                  ]]>
                           </EquationSource>
                        </InlineEquation>
                        collected with the ATLAS detector at the LHC in 2015, in which more than 700 event classes and more than
                        <InlineEquation ID="IEq3">
                           <InlineMediaObject>
                              <ImageObject Color="BlackWhite" FileRef="10052_2019_6540_Article_IEq3.gif" Format="GIF" Rendition="HTML" Type="Linedraw" />
                           </InlineMediaObject>
                           <EquationSource Format="TEX">
                              <![CDATA[$$10^5$$]]>
                           </EquationSource>
                           <EquationSource Format="MATHML">
                              <![CDATA[
                    ]]>
                              <math xmlns:xlink="http://www.w3.org/1999/xlink">
                                 <msup xmlns:xlink="http://www.w3.org/1999/xlink">
                                    <mn xmlns:xlink="http://www.w3.org/1999/xlink">10</mn>
                                    <mn xmlns:xlink="http://www.w3.org/1999/xlink">5</mn>
                                 </msup>
                              </math>
                              <![CDATA[
                  ]]>
                           </EquationSource>
                        </InlineEquation>
                        regions have been analysed. No significant deviations are found and consequently no data-derived signal regions for a follow-up analysis have been defined.
                     </Para>
                  ```

Comment (Neodia) [reply dags/springer/parser.py:112]: @drjova There is some inner tags that we don't need. The xml lib that I'm using doesn't handle those cases so that's why I need to do all this filtering

File: dags/springer/parser.py (hunk 3 above)
Comment (Neodia) [reply dags/springer/parser.py]: Can this happen ? If so, how should we process them ? Currently it will just throw an exception, as it should Imo

File: tests/integration/springer/test_dag_process_file.py (hunk 4 above)
Comment (Neodia) [reply tests/integration/springer/test_dag_process_file.py:48]: This is a bad idea as we are creating an ID for the dagrun. If we create a dagrun with the same id Airflow doesn't like it and throws and error

File: tests/units/springer/test_parser.py (hunk 5 above)
Comment (Neodia) [reply tests/units/springer/test_parser.py]: Good question, I copied pasted without thinking much apparently 🙃 
Gonna remove this

File: dags/springer/parser.py (hunk 2 above)
Comment (Neodia) [reply dags/springer/parser.py]: Removed

File: tests/units/springer/test_parser.py (hunk 5 above)
Comment (Neodia) [reply tests/units/springer/test_parser.py]: Removed

File: dags/springer/dag_process_file.py (hunk 6 above)
Comment (Neodia) [reply dags/springer/dag_process_file.py]: Done

File: dags/springer/parser.py [hunk 8]
```diff
@@ -0,0 +1,256 @@
...
+            return [
+                {
+                    "license": "CC-" + license_type.upper() + "-" + version,
+                    "url": f"{text}/{license_type}/{version}",
+                }
+            ]
+
+        # return default licence if not found
+        # logger.warning(
+        #     'Licence not found, returning default licence for article %s.' % dois)
```
drjova: leftover?

dags/springer/parser.py
drjova: It's not a required field that's why I would just log and not fail the whole harvest.

tests/integration/springer/test_dag_process_file.py
drjova: Good point!

dags/springer/parser.py
Neodia: Removed

dags/springer/parser.py
Neodia: About that, I currently don't have proper logging, should I add it now ? On your leftover comments I removed the comments, but maybe I should just add the logging
$

drjova (dismissed):
LGTM 🚀

drjova:
@Neodia some tests are failing though, could you please have a look?

dags/springer/parser.py
drjova: Yes we should logging, could you make a separate PR? We can use https://github.com/hynek/structlog to give structure in the logs having information like `doi`, `file` to easily filter logs

PR #10, opened 2022-05-02
APS: Fetch from api

ErnestaP:


Makefile
```diff
@@ -43,4 +43,10 @@ stop:
 	cat $(WORKER_PID) | xargs kill
 	cat $(FLOWER_PID) | xargs kill
 	rm *.out *.err *.log
+	kill -9 $(lsof -ti:8080)
+	rm *.out *.err
 	echo -e "\033[0;32m Airflow Stoped. \033[0m"
+
+clean:
+	docker-compose down --volumes --remove-orphans
+	docker rmi $(docker images -a -q) --force
```
Neodia: I think this should be removed. If you want to force new images you could add `--build` when doing `docker-compose up`

dags/aps/APSParams.py
```diff
@@ -0,0 +1,21 @@
+class APSParams:
+    def __init__(
+        self,
+        from_date: str = '2000-01-01',
+        until_date: str = '2000-01-02',
```
Neodia: I would put `from_date = today - 1 day` and `until_date = today` instead of hard coded strings

dags/common/request.py
```diff
@@ -0,0 +1,53 @@
...
+        base_url: str = 'localhost',
+        port: int = 8080,
+        parameters: dict = {},
+        subdir: str ='',
+        headers: dict = {}
+    ):
+        self.base_url = base_url 
+        self.parameters = parameters
+        self.headers = headers
+        self.subdir = subdir
```
Neodia: What is subdir exactly ?

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,30 @@
+from os import getenv
+import json
+
+from common.request import Request
+
+class APSApiClient(Request):
```
Neodia: The class looks great to me ! I would just change the relationship between the class and the `Request` one : The API client is not a Request, it uses requests. Let's take "get_articles_metadata" for example : 

```python
def get_articles_metadata(self):
  request = Request(self.base_url, self.port, parameters, headers)
  response_content = request.get_response_content()
  ...
```

dags/common/request.py
```diff
@@ -0,0 +1,53 @@
...
+        return response
+
+    def get_response_content(self):
+        response = self.get_response()
+        if response.status_code == 200:
+            return response.content
+        print(f'Error while trying to make a request, status code {response.status_code}')
+        raise RequestException
+
+    def set(self, base_url="", parameters={}, headers={}):
```
Neodia: This should be removed. If we need to change something it means we're running a new request => new instance of the object

drjova:
@ErnestaP a general comment, have you install `pre-commit`? Your PR's code is not formatted properly

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,40 @@
...
+            return response_content
+        return ''
+
+    def get_pdf_file(self, doi):
+        request = Request(base_url=self.base_url, parameters=self.parameters)
+        headers = {'Accept': 'application/pdf'}
+        self.subdir = doi
+        response_content = request.get_response_content()
+        request.base_url = self.base_url
+        request.headers = headers
```
drjova: why setting the `base_url` and `headers` after the request?

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,40 @@
...
+        return response_content
+
+    def get_xml_file(self, doi):
+        request = Request(base_url=self.base_url, parameters=self.parameters)
+        headers = {'Accept': 'text/xml'}
+        self.subdir = doi
+        response_content = request.get_response_content()
+        request.base_url = self.base_url
+        request.headers = headers
```
drjova: ditto

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,50 @@
...
+    repo.delete_all()
+    today = date.today()
+    aps_fetching_from_date = repo.find_the_last_uploaded_file_date()
+    aps_fetching_until_date = today.strftime("%Y-%m-%d")
+    if aps_fetching_from_date is None:
+        Variable.set('aps_fetching_from_date', aps_fetching_until_date)
+    else:
+        Variable.set('aps_fetching_from_date', aps_fetching_from_date)
+    Variable.set('aps_fetching_until_date', aps_fetching_until_date)
+    return({'aps_fetching_from_date': Variable.get('aps_fetching_from_date'), 'aps_fetching_until_date': Variable.get('aps_fetching_until_date')})
```
drjova: Why we have to `set`? We just return the values, no?

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,40 @@
...
+import json
+
+from common.request import Request
+from aps.APSParams import APSParams
+
+
+class APSApiClient:
+    def __init__(self, base_url=getenv("BASE_URL_FOR_APS", 'http://harvest.aps.org/v2/journals/articles'), parameters=APSParams()):
+        self.base_url = base_url
+        self.parameters = parameters
```
Neodia: This should be removed

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def save_file_in_s3(rest_api, repo):
+    articles_metadata = rest_api.get_articles_metadata()
+    if bool(articles_metadata):
+        date_today =  date.today().strftime("%Y-%m-%d")
+        prefix = date.today().strftime("%Y-%m-%d_%H:%M:%S")
+
+        byte_file = io.BytesIO(articles_metadata)
+        key = '/'.join([date_today, '.'.join([prefix, 'json'])])
```
drjova: Easier to read ☺️ 
```suggestion
   key = '{date_today}/{prefix}.json'
```

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,40 @@
...
+        response_content = request.get_response_content()
+        json_obj = json.loads(response_content)
+        if len(json_obj['data']) > 0:
+            return response_content
+        return ''
+
+    def get_pdf_file(self, doi):
+        request = Request(base_url=self.base_url, parameters=self.parameters)
+        headers = {'Accept': 'application/pdf'}
+        self.subdir = doi
```
Neodia: This should be removed. We can discuss it after if you want

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,50 @@
...
+        Variable.set('aps_fetching_from_date', aps_fetching_from_date)
+    Variable.set('aps_fetching_until_date', aps_fetching_until_date)
+    return({'aps_fetching_from_date': Variable.get('aps_fetching_from_date'), 'aps_fetching_until_date': Variable.get('aps_fetching_until_date')})
+
+
+def save_file_in_s3(rest_api, repo):
+    articles_metadata = rest_api.get_articles_metadata()
+    if bool(articles_metadata):
+        date_today =  date.today().strftime("%Y-%m-%d")
+        prefix = date.today().strftime("%Y-%m-%d_%H:%M:%S")
```
drjova: This will return something like:

```
2022-01-01_2022-01-01_12:12:12
```

Why do we want this? should it be only the prefix enough?

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,50 @@
...
+
+        byte_file = io.BytesIO(articles_metadata)
+        key = '/'.join([date_today, '.'.join([prefix, 'json'])])
+        repo.save(key, byte_file)
+        return key
+        
+def split_json(repo, key):
+    file = repo.find_by_id(key)
+    data = json.loads(file.getvalue().decode('utf-8'))['data']
```
drjova: why we need to decode?

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,50 @@
...
+    ids_and_articles=[]
+    for article in data:
+        doi = article['identifiers']['doi']
+        today =  datetime.now().strftime('%Y-%m-%dT%H:%M:%S.%f')
+        id = '_'.join([f'APS_{doi}_', today])
```
drjova: This is wrong, it will return 
```APS_DOI__2022-01-01:XXXXXXX```

```suggestion
        id = f"APS_{doi}_{today}"
```

dags/aps/repository.py
```diff
@@ -0,0 +1,38 @@
...
+        return retfile
+
+    def find_the_last_uploaded_file_date(self):
+        dates = []
+        for obj in list(self.s3_bucket.objects.all()):
+            dates.append(obj.last_modified.strftime("%Y-%m-%d"))
+        if len(dates) == 0:
+            return None
+        the_newest_file_date = max(dates)
+        return the_newest_file_date
```
drjova:
```suggestion
        objects = list(self.s3_bucket.objects.all())
        if not objects:
            return 
        dates = [obj.last_modified.strftime("%Y-%m-%d") for obj in objects]
        return max(dates)
```

dags/common/request.py
```diff
@@ -0,0 +1,45 @@
...
+
+        url_base_obj = furl(self.base_url)
+        if bool(self.subdir):
+            url_base_obj = url_base_obj.path(self.subdir)
+        if bool(self.parameters):
+            url_base_obj = url_base_obj.add(self.parameters)
+        if bool(self.headers):
+            response = requests.get(url_base_obj.url, headers=self.headers)
+            return response
+        response=requests.get(url_base_obj.url)
```
drjova:
```suggestion
        response = requests.get(url_base_obj.url)
```

drjova (changes requested):

dags/aps/fetch_from_api.py
Neodia: If I understand correctly `loads` wants a String. `getvalue` returns bytes so we must decode them to get a String

dags/common/pull_ftp.py
```diff
@@ -9,14 +9,15 @@ def migrate_from_ftp(sftp: SFTPService, repo: IRepository):
...
+            continue
         with zipfile.ZipFile(file_bytes) as zip:
-            for zip_filename in zip.namelist():
-                file_prefix = file.split(".")[0]
-                file_content = zip.read(zip_filename)
-                repo.save(file_prefix + "/" + zip_filename, io.BytesIO(file_content))
+                    for zip_filename in zip.namelist():
+                        file_prefix = file.split(".")[0]
+                        file_content = zip.read(zip_filename)
+                        repo.save(file_prefix + "/" + zip_filename, io.BytesIO(file_content))
```
drjova: it has some weird spacing and wrong indentation

tests/integration/common/test_rest_api_service.py
```diff
@@ -0,0 +1,24 @@
...
+@pytest.fixture
+def rest_api_with_params_fixture(rest_api_fixture):
+    rest_api = rest_api_fixture
+    assert exptected_params == rest_api.get()
+    return rest_api
+
+def test_get_response(rest_api_with_params_fixture: Request):
+    reponse = rest_api_with_params_fixture.get_response()
+    assert reponse.status_code == 200
+
```
drjova: could you please configure your editor to add a new line at the end of each file?

dags/common/request.py
```diff
@@ -0,0 +1,45 @@
...
+    def __repr__(self) -> str:
+        return(f"< RestApi base_url={self.base_url}, parameters={self.parameters}, header={self.headers} >")
+        
+    def get_response(self):
+        from furl import furl
+        import requests
+
+        url_base_obj = furl(self.base_url)
+        if bool(self.subdir):
+            url_base_obj = url_base_obj.path(self.subdir)
```
drjova: this doesn't work, it should be `url_base_obj.path = self.subdir`

dags/common/request.py
drjova: you don't have to do the `if`s since the library will take care it if the value is `None`

dags/common/request.py
drjova: Don't we have tests for that? It should have caught in the tests 🤔

drjova (changes requested):

dags/aps/fetch_from_api.py
ErnestaP: it's like that:
2022-01-01/2022-01-01_12:12:12.json

dags/common/request.py
```diff
@@ -23,14 +23,10 @@ def get_response(self):
...
-            url_base_obj = url_base_obj.path(self.subdir)
-        if bool(self.parameters):
-            url_base_obj = url_base_obj.add(self.parameters)
-        if bool(self.headers):
-            response = requests.get(url_base_obj.url, headers=self.headers)
-            return response
-        response = requests.get(url_base_obj.url)
+        url_base_obj.path.segments = self.path_segments
+        url_base_obj.add(self.parameters)
+        print(url_base_obj.url)
```
drjova: I would remove all the prints and just add logs

tests/integration/aps/test_fetch_from_api.py
```diff
@@ -0,0 +1,98 @@
...
+    save_file_in_s3(aps_api_client, repo, parameters)
+    assert len(repo.find_all()) == 1
+
+
+def __test_dagrun_state(dagrun: DagRun):
+    dagrun.update_state()
+    return (
+        not dagrun.get_state() == DagRunState.QUEUED
+        and not dagrun.get_state() == DagRunState.RUNNING
+    )
```
drjova, line 84: any particular reason why `__`?

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,62 @@
...
+        path_segments=None,
+    ):
+        self.base_url = (
+            getenv("APS_BASE_URL", "http://harvest.aps.org")
+            if base_url is None
+            else base_url
+        )
+        self.path_segments = (
+            ["v2", "journals", "articles"] if path_segments is None else path_segments
+        )
```
drjova:
```suggestion
        self.path_segments = path_segments or  ["v2", "journals", "articles"]
```
☺️

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,62 @@
...
+        request = Request(
+            base_url=self.base_url,
+            path_segments=all_path_segments,
+            parameters=parameters,
+        )
+        response_content = request.get_response_content()
+        json_obj = json.loads(response_content)
+        if len(json_obj["data"]) > 0:
+            return response_content
+        return ""
```
drjova: here it returns empty string, why not `None`?

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,62 @@
...
+            base_url=self.base_url,
+            path_segments=all_path_segments,
+            parameters=parameters,
+            headers=headers,
+        )
+        response_content = request.get_response_content()
+        return response_content
+
+    def __repr__(self) -> str:
+        return f"< APSApiClient base_url={self.base_url}, path_segments={self.path_segments}, parameters={self.parameters} >"
```
drjova: this should be wrapped in multiple lines :)

dags/common/request.py
```diff
@@ -0,0 +1,49 @@
...
+        headers: dict = {},
+        path_segments: list = [],
+    ):
+        self.base_url = base_url
+        self.parameters = parameters
+        self.headers = headers
+        self.path_segments = path_segments
+
+    def __repr__(self) -> str:
+        return f"< RestApi base_url={self.base_url}, parameters={self.parameters}, header={self.headers}, path_segments={self.path_segments} >"
```
drjova: same this should be wrapped in multiple lines

dags/common/request.py
```diff
@@ -0,0 +1,49 @@
...
+        url_base_obj.path.segments = self.path_segments
+        url_base_obj.add(self.parameters)
+        logging.debug(url_base_obj.url)
+        response = requests.get(url_base_obj.url, headers=self.headers)
+        return response
+
+    def get_response_content(self):
+        response = self.get_response()
+        if response.status_code == 200:
+            return response.content
```
drjova, line 36: we are expecting json here, right? Why not `response.json()` to avoid every time to run ` json.loads`?

tests/integration/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_get_pdf_file(aps_api_client_fixture):
+    doi = "10.1103/PhysRevLett.126.153601"
+    parameters = APSParams(from_date="2021-04-10", until_date="2022-04-10").get_params()
+    pdf = aps_api_client_fixture.get_pdf_file(parameters=parameters, doi=doi)
+    data_folder = os.path.join(os.getcwd(), "data")
+    file_path = os.path.join(data_folder, "test_pdf.pdf")
+    with open(file_path, "rb") as pdf_file:
+        assert pdf_file.read() == pdf
```
drjova: we should use https://pypi.org/project/pytest-datadir/ :)

tests/integration/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,44 @@
...
+    ).get_params()
+    metadata_value = aps_api_client_fixture.get_articles_metadata(parameters=parameters)
+    assert type(metadata_value) == type(expected_metadata_value)
+    assert metadata_value == expected_metadata_value
+
+
+def test_get_pdf_file(aps_api_client_fixture):
+    doi = "10.1103/PhysRevLett.126.153601"
+    parameters = APSParams(from_date="2021-04-10", until_date="2022-04-10").get_params()
+    pdf = aps_api_client_fixture.get_pdf_file(parameters=parameters, doi=doi)
```
drjova, line 30: this test runs directly into the APS server, we shouldn't do that because:
* if the APS goes down our tests are relying on that and will fail as well
* We will create unnecessary traffic to their server

Hence, we should mock the response from APS. To do that you should use https://vcrpy.readthedocs.io/en/latest/, let me know if you would like to discuss it IRL

This applies to all tests that are getting data from APS servers

tests/integration/aps/test_fetch_from_api.py
```diff
@@ -0,0 +1,109 @@
...
+    dagrun = dag.create_dagrun(DagRunState.QUEUED, run_id=id)
+    wait().at_most(60, SECOND).until(lambda: __test_dagrun_state(dagrun))
+
+    queued_tasks = get_dag_runs_amount(TRIGGERED_DAG_NAME, state="queued")
+    running_tasks = get_dag_runs_amount(TRIGGERED_DAG_NAME, state="running")
+    queued_and_running_tasks = queued_tasks + running_tasks
+    if len(repo.find_all()) == 0:
+        assert 0 == queued_and_running_tasks
+    else:
+        assert queued_and_running_tasks > 0
```
drjova: I don't understand this test as it is not deterministic. One of the two cases should be true, why is it possible to have both cases?

dags/common/request.py
ErnestaP: tests/units/common/test_request.py - tests added here

tests/integration/aps/test_fetch_from_api.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def test_dag_run(dag: DAG):
+    repo = APSRepository()
+    repo.delete_all()
+    assert len(repo.find_all()) == 0
+    id = datetime.datetime.utcnow().strftime("test_aps_fetch_api_%Y-%m-%dT%H:%M:%S.%f")
+    dagrun = dag.create_dagrun(DagRunState.QUEUED, run_id=id)
+    wait().at_most(60, SECOND).until(lambda: __test_dagrun_state(dagrun))
+    assert len(repo.find_all()) == 0 or len(repo.find_all()) == 1
```
drjova: same, one of the two should be correct.

tests/units/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,45 @@
...
+def mocked_empty_response(*args, **kwrgs):
+    mocked_response = mock.Mock()
+    mocked_response.status_code = 200
+    mocked_response.url = args[0]
+    mocked_response.content = '{"data":[]}'
+    return mocked_response
+
+
+@mock.patch("common.request.requests.get", side_effect=mocked_empty_response)
+def test_get_articles_metada_empty(mocked_response):
```
drjova:
```suggestion
def test_get_articles_metadata_empty(mocked_response):
```

tests/units/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,45 @@
...
+def mocked_response(*args, **kwrgs):
+    mocked_response = mock.Mock()
+    mocked_response.status_code = 200
+    mocked_response.url = args[0]
+    mocked_response.content = '{"data":[{"abstract":{"value":"<p>We propose and theoretically analyze</p>"}}]}'
+    return mocked_response
+
+
+@mock.patch("common.request.requests.get", side_effect=mocked_response)
+def test_get_articles_metada(mocked_response):
```
drjova:
```suggestion
def test_get_articles_metadata(mocked_response):
```

dags/aps/APSApiClient.py
Neodia: Wow so cool !

tests/units/aps/test_fetch_from_api.py
```diff
@@ -0,0 +1,58 @@
...
+    dates = set_APS_harvesting_interval(repo)
+    assert dates == expected_days
+
+
+def test_save_file_in_s3(
+    rest_api=MockedAPSApiClient(),
+    repo=MockedRepo(),
+    parameters=APSParams(expected_days).get_params(),
+):
+    exptected_key = f'{today.strftime("%Y-%m-%d")}/{ datetime.now().strftime("%Y-%m-%dT%H:%M")}.json'
```
drjova: this test will be flaky, instead you could "hardcode" the dates using https://github.com/spulec/freezegun this will help the test to always run with the same date and time and not rely on `datetime.now()` which will be different in every run

dags/common/request.py
Neodia: Don't we have a pre-commit plugin to detect line that are too long ?

tests/units/aps/test_trigger_files_processing.py
```diff
@@ -0,0 +1,76 @@
...
+        runs_ids = get_dag_runs_amount(
+            TRIGGERED_DAG_NAME, states=["success", "running", "queued", "failed"]
+        )
+    except DagRunAlreadyExists:
+        runs_ids = get_dag_runs_amount(
+            TRIGGERED_DAG_NAME, states=["success", "running", "queued", "failed"]
+        )
+    finally:
+        for obj in id_and_articles:
+            assert obj["id"] in runs_ids
```
drjova: here we have two cases and I don't understand how the `DagRunAlreadyExists` raises? is it because of a previous test?

drjova (commented):
@ErnestaP looks good, please have a look on the comments and then it's ready 🚀

dags/common/request.py
ErnestaP: it passes pre-commit :/

tests/integration/aps/test_fetch_from_api.py
ErnestaP: yes, it can be both, because if there was no articles uploaded today, no json will be saved on S3

dags/common/request.py
drjova: probably because you edit this file before you install pre-commit, try to modify the file and commit, it should fix it

tests/units/aps/test_trigger_files_processing.py
ErnestaP: yes, and I didn't find how to clean runs from the DAG

tests/integration/aps/test_fetch_from_api.py
drjova: This is wrong, please check my comment https://github.com/cern-sis/workflows/pull/10#discussion_r864829769

ErnestaP: it's possible because is not on us: if there was no articles uploaded to day, no json will be downloaded

tests/integration/aps/test_aps_api_client.py
ErnestaP: there are tests with mocked response here: tests/units/common/test_request.py
is it okey?

drjova: yes there are tests with mocked response, but these tests here are accessing directly the APS server, this is not a good practice unfortunately.

ErnestaP: so integration test has to mock response as well?

drjova: yes, any external service should be mocked. Please use `vcrpy` to do that

tests/integration/aps/test_fetch_from_api.py
drjova: That's why we have to have a specific date and mock the response. The tests should be deterministic in order to make sense.

tests/units/aps/test_trigger_files_processing.py
drjova: You can create a fixture which deletes all DAGS after each test :)

.gitignore
```diff
@@ -58,7 +58,7 @@ coverage.xml
 *.py,cover
 .hypothesis/
 .pytest_cache/
-
+tests/integration/aps/cassettes/
```
drjova: why you ignore this 😅 ? This should be committed in order to run the tests with these cassettes

dags/aps/APSApiClient.py
```diff
@@ -0,0 +1,55 @@
...
+            parameters=parameters,
+            headers=headers,
+        )
+        response_content = request.get_response_bytes()
+        return response_content
+
+    def __repr__(self) -> str:
+        return f"< APSApiClient base_url={self.base_url}, \
+                    path_segments={self.path_segments}, \
+                    parameters={self.parameters} >"
```
drjova:
```suggestion
        return f"<APSApiClient base_url={self.base_url}, \
                    path_segments={self.path_segments}, \
                    parameters={self.parameters}>"
```

dags/aps/aps.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+        return save_file_in_s3(rest_api=rest_api, repo=repo, parameters=parameters)
+
+    @task()
+    def trigger_file_processing(key, repo: IRepository = APSRepository()):
+        if key is None:
+            warnings.warn("No new files were downloaded to s3")
+        else:
+            ids_and_articles = split_json(repo, key)
+            trigger_file_processing_(ids_and_articles)
```
drjova:
```suggestion
        if key is None:
            warnings.warn("No new files were downloaded to s3")
            return
        ids_and_articles = split_json(repo, key)
        trigger_file_processing_(ids_and_articles)
```

Why are you using `warnings` instead of logging? 🤔

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def set_APS_harvesting_interval(repo):
+    today = date.today()
+    aps_fetching_from_date = repo.find_the_last_uploaded_file_date()
+    aps_fetching_until_date = today.strftime("%Y-%m-%d")
+    if aps_fetching_from_date is None:
+        Variable.set("aps_fetching_from_date", aps_fetching_until_date)
+    else:
+        Variable.set("aps_fetching_from_date", aps_fetching_from_date)
```
drjova:
```suggestion
    Variable.set("aps_fetching_from_date", aps_fetching_until_date or aps_fetching_from_date)
```

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,61 @@
...
+    Variable.set("aps_fetching_until_date", aps_fetching_until_date)
+    return {
+        "aps_fetching_from_date": Variable.get("aps_fetching_from_date"),
+        "aps_fetching_until_date": Variable.get("aps_fetching_until_date"),
+    }
+
+
+def save_file_in_s3(rest_api, repo, parameters):
+    #  io.BytesIO(json.dumps(pseudo_article).encode("utf-8"))
+    # print(type(rest_api.get_articles_metadata(parameters)))
```
drjova: leftovers?

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,61 @@
...
+    articles_metadata = str.encode(
+        json.dumps(rest_api.get_articles_metadata(parameters))
+    )
+    print(type(articles_metadata))
```
drjova: We don't need the `print`

dags/aps/aps.py
```diff
@@ -0,0 +1,44 @@
+import os
+import warnings
+
+import airflow
+from airflow.decorators import dag, task
+from aps.APSApiClient import APSApiClient
```
drjova:
```suggestion
from aps.aps_api_client import APSApiClient
```
We have to rename the filename

dags/aps/aps.py
```diff
@@ -0,0 +1,44 @@
...
+from aps.APSParams import APSParams
```
drjova:
```suggestion
from aps.aps_params import APSParams
```

dags/aps/APSParams.py
```diff
@@ -0,0 +1,31 @@
...
+        params = {
+            "from": self.from_date,
+            "until": self.until_date,
+            "date": self.date,
+            "journals": self.journals,
+            "sets": self.sets,
+            "per_page": self.per_page,
+        }
+        cleaned_parameters = {key: value for key, value in params.items() if value}
+        return cleaned_parameters
```
drjova:
```suggestion
        return {key: value for key, value in params.items() if value}
```

dags/aps/aps.py
```diff
@@ -0,0 +1,44 @@
...
+    @task()
+    def save_json_in_s3(dates: dict, repo: IRepository = APSRepository()):
+        parameters = APSParams(
+            from_date=dates["aps_fetching_from_date"],
+            until_date=dates["aps_fetching_until_date"],
+        ).get_params()
+        rest_api = APSApiClient(
+            base_url=os.getenv("APS_BASE_URL", "http://harvest.aps.org")
+        )
+        return save_file_in_s3(rest_api=rest_api, repo=repo, parameters=parameters)
```
drjova: the logic to fetch the JSON from the rest api should belong here. Since the `save_file_in_s3` is an aps function and not a common function, it shouldn't have the logic to interact with the rest API, the sole responsibility of this function is to save the file ([more details](https://en.wikipedia.org/wiki/Single-responsibility_principle)).

dags/aps/aps.py
```diff
@@ -0,0 +1,44 @@
...
+from airflow.decorators import dag, task
+from aps.APSApiClient import APSApiClient
+from aps.APSParams import APSParams
+from aps.fetch_from_api import save_file_in_s3, set_APS_harvesting_interval, split_json
+from aps.fetch_from_api import trigger_file_processing as trigger_file_processing_
+from aps.repository import APSRepository
+from common.repository import IRepository
+
+
+@dag(start_date=airflow.utils.dates.days_ago(0))
```
drjova, line 15: Do we have to specify this?

dags/aps/fetch_from_api.py
```diff
@@ -0,0 +1,61 @@
+import io
```
drjova: these functions are `utils` so we can name the module as `utils.py`

dags/common/request.py
```diff
@@ -0,0 +1,64 @@
...
+            f"Error while trying to make a request, status code {response.status_code}"
+        )
+        raise requests.exceptions.RequestException
+
+    def get_response_json(self):
+        import json
+
+        response = self.get_response()
+        if response.status_code == 200:
+            return json.loads(response.content)
```
drjova: `return response.json`

dags/common/request.py
```diff
@@ -0,0 +1,64 @@
...
+                    path_segments={self.path_segments} >"
+
+    def get_response(self):
+        from furl import furl
+
+        url_base_obj = furl(self.base_url)
+        url_base_obj.path.segments = self.path_segments
+        url_base_obj.add(self.parameters)
+        logging.debug(url_base_obj.url)
+        response = requests.get(url_base_obj.url, headers=self.headers)
```
drjova, line 33:
```suggestion
        response = requests.get(url_base_obj.url, headers=self.headers)
        response.raise_for_status()
```

dags/common/request.py
```diff
@@ -0,0 +1,64 @@
...
+        return response
+
+    def get_response_bytes(self):
+        response = self.get_response()
+        print(response.content)
+        if response.status_code == 200:
```
drjova: since you will add `raise_for_status()` no need to check the status

dags/common/request.py
```diff
@@ -0,0 +1,64 @@
...
+        logging.error(
+            f"Error while trying to make a request, status code {response.status_code}"
+        )
+        raise requests.exceptions.RequestException
+
+    def get_response_json(self):
+        import json
+
+        response = self.get_response()
+        if response.status_code == 200:
```
drjova: since you will add `raise_for_status()` no need to check the status

dags/common/request.py
```diff
@@ -0,0 +1,64 @@
...
+        self.base_url = base_url
+        self.parameters = parameters
+        self.headers = headers
+        self.path_segments = path_segments
+
+    def __repr__(self) -> str:
+        return f"< RestApi base_url={self.base_url}, \
+                    parameters={self.parameters}, \
+                    header={self.headers}, \
+                    path_segments={self.path_segments} >"
```
drjova:
```suggestion
        return f"<RestApi base_url={self.base_url}, \
                    parameters={self.parameters}, \
                    header={self.headers}, \
                    path_segments={self.path_segments}>"
```

dags/common/request.py
```diff
@@ -0,0 +1,64 @@
...
+
+        response = self.get_response()
+        if response.status_code == 200:
+            return json.loads(response.content)
+        logging.error(
+            f"Error while trying to make a request, status code {response.status_code}"
+        )
+        raise requests.exceptions.RequestException
+
+    def get(self):
```
drjova: `get` is confusing in a class named `Request`, I would expect to do a `GET` request instead of getting parameters. Maybe `get_parameters` is more relevant? WDYT?

dags/common/request.py
```diff
@@ -0,0 +1,64 @@
...
+        if bool(self.headers):
+            return {
+                "base_url": self.base_url,
+                "parameters": self.parameters,
+                "headers": self.headers,
+            }
+        return {"base_url": self.base_url, "parameters": self.parameters}
```
drjova: To avoid rewriting the same statements:
```suggestion
        parameters = {
            "base_url": self.base_url,
            "parameters": self.parameters,
        }
        if bool(self.headers):
            parameters['headers'] = self.headers
        return parameters
```

tests/integration/aps/test_fetch_from_api.py
```diff
@@ -0,0 +1,78 @@
...
+    assert dagbag.import_errors.get(f"dags/{DAG_NAME}.py") is None
+    return dagbag.get_dag(dag_id=DAG_NAME)
+
+
+def test_dag_loaded(dag: DAG):
+    assert dag is not None
+    assert len(dag.tasks) == 3
+
+
+@pytest.mark.vcr(record_mode="new")
```
drjova: I don't see any `new` mode in the [docs](https://vcrpy.readthedocs.io/en/latest/usage.html#once), we should use `once` instead.

dags/aps/aps.py
ErnestaP: because it is a warnings  that no new files were downloaded, so no processes will be triggered

ErnestaP: I don't understand the logic of the suggestion 😅

drjova: why not using `logging.warn`?

The logic of the suggestion is to avoid having `if else`

dags/aps/APSParams.py
ErnestaP: why is bad to create a var. with a name to make it more clear?

dags/aps/aps.py
ErnestaP: yes:
1. is mandatory to have starting date
2. On order to start dag immediately

dags/aps/APSParams.py
drjova: we are not using this var anywhere, hence we don't need to story it

Neodia: This being the return value of the function, we should know what's being returned by reading the function so no need for extra var for readability

tests/integration/aps/test_aps_api_client.py
drjova: This still doesn't have a cassette

.gitignore
drjova: we should remove it from here.

tests/integration/aps/test_fetch_from_api.py
```diff
@@ -0,0 +1,80 @@
...
+    assert dagbag.import_errors.get(f"dags/{DAG_NAME}.py") is None
+    return dagbag.get_dag(dag_id=DAG_NAME)
+
+
+def test_dag_loaded(dag: DAG):
+    assert dag is not None
+    assert len(dag.tasks) == 3
+
+
+@pytest.mark.vcr(record_mode="new_episode")
```
drjova: why `new_episode`? Didn't we already discussed it to add `once`?

> I don't see any `new` mode in the [docs](https://vcrpy.readthedocs.io/en/latest/usage.html#once), we should use `once` instead.

tests/integration/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,43 @@
...
...0-959c-4e90-ba0e-38232980802a"},"primary":false},{"id":"2c0e953d-44fd-4611-bfb0-3e39cfb9374a","facet":{"id":"bdb1ef91-b776-4e36-8f8f-3e93666bac1e"},"primary":false},{"id":"1fa8daba-8a48-4d9f-ada9-03a9d3259ca3","facet":{"id":"bdb1ef91-b776-4e36-8f8f-3e93666bac1e"},"primary":false},{"id":"9b6aaf50-04bd-4aaa-b4c3-d6fa9b7f7303","facet":{"id":"bdb1ef91-b776-4e36-8f8f-3e93666bac1e"},"primary":true},{"id":"632f74d1-d16b-4b92-a295-b24729a93d8d","facet":{"id":"bdb1ef91-b776-4e36-8f8f-3e93666bac1e"},"primary":false}],"disciplines":["510fc218-8774-4547-ab09-fc6cef0c9a03","8ab3d2b7-b50b-460e-b086-8990cfcbd350","a48f173e-6459-4642-a711-a6e731807625"]}},"publisher":{"name":"APS"},"rights":{"rightsStatement":"©2021 American Physical Society","copyrightYear":2021,"copyrightHolders":[{"name":"American Physical Society"}],"creativeCommons":false,"licenses":[{"url":"http://link.aps.org/licenses/aps-default-license"}]},"journal":{"id":"PRA","abbreviatedName":"Phys. Rev. A","name":"Physical Review A"},"title":{"value":"Continuous real-time sensing with a nitrogen-vacancy center via coherent population trapping","format":"html"},"tocSection":{"label":"Quantum technologies"},"volume":{"number":"103"}}]}'
```
drjova: why not a JSON file in `data`?

tests/integration/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,43 @@
...
+)
+
+
+def test_get_articles_metadata(aps_api_client_fixture):
```
drjova: aren't these tests querying APS?

tests/units/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,53 @@
...
+    mocked_response.url = args[0]
+    mocked_response.content = '{"data":[{"abstract":{"value":"<p>We propose and theoretically analyze</p>"}}]}'
+    mocked_response.json = mock.MagicMock(
+        return_value={
+            "data": [
+                {"abstract": {"value": "<p>We propose and theoretically analyze</p>"}}
+            ]
+        }
+    )
+    # mocked_response.json() = {"data":[{"abstract":{"value":"<p>We propose and theoretically analyze</p>"}}]}
```
drjova: you don't need this

tests/units/aps/test_utils.py
```diff
@@ -0,0 +1,76 @@
...
+        today = date.today()
+        return today
+
+    def save(self, key, file):
+        pass
+
+    def find_by_id(self, id):
+        return BytesIO(
+            str.encode(
+                '{"data":[{"abstract":{"value":"<p>We propose and theoretically analyze</p>"},  "identifiers":{"doi":"10.1103/PhysRevLett.126.153601"}}]}'
```
drjova: this line is too long, did you run pre-commit?

drjova (changes requested):
Few more comments and looks good!

Reading the [vcrpy docs](https://vcrpy.readthedocs.io/en/latest/usage.html?highlight=mode#once) once is the correct mode, which means there is something wrong with the tests. Have you compared the cassettes between two runs? The problem is with the `Headers` and `date` since it changes every time the test runs. You should use freeze gun for the dates and ignore the headers such as `X-Amz-Date` etc.

An example to ignore headers is the following 

```python
@pytest.fixture(scope="session")
def vcr_config():
    return {
        "ignore_localhost": True,
        "decode_compressed_response": True,
        "filter_headers": ("Authorization", "User-Agent"),
        "record_mode": "once",
    }
```

tests/units/aps/test_utils.py
ErnestaP: yes :(

tests/integration/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+@pytest.fixture(scope="session")
+def vcr_config():
+    return {
+        "ignore_localhost": True,
+        "decode_compressed_response": True,
+        "filter_headers": ("X-Amz-Date"),
+        "record_mode": "once",
+    }
```
drjova: if you add this here is just for this file, I would suggest to have it globally in the `tests/integration/conftest.py` file :)

tests/integration/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,53 @@
...
+def test_get_pdf_file(aps_api_client_fixture, shared_datadir):
+    doi = "10.1103/PhysRevLett.126.153601"
+    parameters = APSParams(from_date="2021-04-10", until_date="2022-04-10").get_params()
+    pdf = aps_api_client_fixture.get_pdf_file(parameters=parameters, doi=doi)
+    file_path = os.path.join(shared_datadir, "test_pdf.pdf")
+    with open(file_path, "rb") as pdf_file:
+        assert pdf_file.read() == pdf
+
+
+@pytest.mark.vcr(record_mode="once")
```
drjova: since you have a global config for `vcrpy` you don't need to specify the `once` in every test, could you please remove it?

tests/integration/aps/test_aps_api_client.py
```diff
@@ -0,0 +1,53 @@
...
+from aps.aps_api_client import APSApiClient
+from aps.aps_params import APSParams
+
+
+@pytest.fixture(scope="session")
+def vcr_config():
+    return {
+        "ignore_localhost": True,
+        "decode_compressed_response": True,
+        "filter_headers": ("X-Amz-Date"),
```
drjova: could you please add `Authorization` as well?

tests/integration/confest.py
```diff
@@ -0,0 +1,12 @@
...
+
+@freeze_time("2022-05-04")
+@pytest.fixture(scope="session")
+def vcr_config():
+    return {
+        "ignore_localhost": True,
+        "decode_compressed_response": True,
+        "filter_headers": ("Authorization", "X-Amz-Date"),
+        "record_mode": "once",
+    }
```
drjova: new line

tests/integration/confest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
+from freezegun import freeze_time
+
+@freeze_time("2022-05-04")
```
drjova: why freeze_time here?

dags/common/request.py
```diff
@@ -0,0 +1,53 @@
...
+        self.base_url = base_url
+        self.parameters = parameters
+        self.headers = headers
+        self.path_segments = path_segments
+
+    def __repr__(self) -> str:
+        return f"RestApi base_url={self.base_url}, \
+                    parameters={self.parameters}, \
+                    header={self.headers}, \
+                    path_segments={self.path_segments}>"
```
drjova:
```suggestion
        return f"<RestApi base_url={self.base_url}, \
                    parameters={self.parameters}, \
                    header={self.headers}, \
                    path_segments={self.path_segments}>"
```

drjova (approved):
LGTM 🚀 please squash!

drjova:
@ErnestaP could you please rebase?

PR #11, opened 2022-05-11
Cleaning: Removing unwanted white spaces

ErnestaP:


tests/units/common/cleaning/test_clean_whitespace_characters.py
```diff
@@ -0,0 +1,19 @@
...
+    output = clean_whitespace_characters(string)
+    assert output == expected_string
+
+def test_clean_whitespace_characters_with_double_spaces():
+    output = clean_whitespace_characters(string_with_double_spaces)
+    assert output == expected_string
+
+def test_clean_whitespace_characters_with_new_lines():
+    output = clean_whitespace_characters(string_with_new_lines)
+    assert output == expected_string
```
drjova: You could use https://docs.pytest.org/en/6.2.x/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions not to repeat the same test :)

Neodia: I personally prefer how @ErnestaP did it there. Tests are not run frequently so I think it's better for them to be readable so we can directly know which is the problem

dags/common/cleaning/clean_whitespace_characters.py
```diff
@@ -0,0 +1,2 @@
+def clean_whitespace_characters(input):
+    return ' '.join(input.split())
```
Neodia: I don't understand how this works with multiple spaces ? Won't this add spaces between spaces ?

For curiosity purposes, any idea on the performance difference with 
```python
re.sub("\\s+", " ", input)
```
?

Neodia:
For the file structure, what about doing `dags/common/cleaning.py` and putting all the cleaning functions there ? I don't think we need a file per function. Same for tests

dags/common/cleaning/clean_whitespace_characters.py
ErnestaP: ```re.sub("\\s+", " ", input)```  
Won't remove the trailing spaces. The expression you suggested will take 1 or more spaces and replace them with one space, as the result will leave the ones at the beginning and end of the string

Neodia: Oh yes good point !

tests/units/common/cleaning/test_clean_whitespace_characters.py
ErnestaP: I see your point and it makes sense. However, I like the suggestion of @drjova , because this test is simple, and does precisely the same thing: compare two strings.  If at least one test fails, we will see the input and expected output in the error

drjova: `parametrise` is the way to go for such cases. It will give you the info to understand which case has the problem by displaying each case as [a separate test](https://docs.pytest.org/en/6.2.x/example/parametrize.html#generating-parameters-combinations-depending-on-command-line). Additionally, you can add new use cases without writing a new test. At the same time, if the function's name changes, you only have to do it in one place and not in `x` places.

Neodia: Ohh yes I see ! It's good that we can put custom names (ids).
Do you mind [adding test ids (names)](https://docs.pytest.org/en/6.2.x/example/parametrize.html#different-options-for-test-ids) to have the best of both approaches ?

Neodia:
Don't forget to rebase on main as I pushed some changes just now

tests/units/common/test_clenup.py
```diff
@@ -0,0 +1,15 @@
+import pytest
```
Neodia: Typo in filename : `clenup` -> `cleanup`

drjova (approved):

Neodia (approved):

PR #12, opened 2022-05-11
Debugging: Added logging for secondary fields

Neodia:


dags/springer/parser.py
```diff
@@ -12,7 +13,7 @@
 
 class SpringerParser(IParser):
     def __init__(self) -> None:
-
+        self.logger = structlog.get_logger()
```
drjova:
```suggestion
        self.logger = get_logger().bind(class_name=type(self).__name__)
```

Neodia: Updated

drjova (approved):

PR #13, opened 2022-05-11
Generic Parsing: Added take_first

Neodia:


dags/common/parsing/generic_parsing.py
```diff
@@ -0,0 +1,4 @@
+def take_first(arr):
+    for value in arr:
+        if value is not None and value != "":
+            return value
```
drjova:
```python
if value: 
   ....
```
is enough to check if it's empty for example:

```python
In [17]: a = ""

In [18]: True if a else False
Out[18]: False

In [19]: b = None

In [20]: True if b else False
Out[20]: False
```

just for reference; a more pythonic way to do this is, it would be something like:

```python
try:
     return next(filter(None, arr))
except StopIteration:
     return None
```

Neodia: Thanks for the suggestion, changed

drjova (approved):

PR #14, opened 2022-05-12
Generic Parsing: Added list_to_value_dict

Neodia:


dags/common/parsing/generic_parsing.py
```diff
@@ -3,3 +3,7 @@ def take_first(arr):
         return next(filter(None, arr))
     except StopIteration:
         return None
+
+
+def list_to_value_dict(arr, key="value"):
+    return [{key: val} for val in arr]
```
drjova: Should we check `if val`? in case the value is empty or None?

Neodia: Do we want {key: None} in our output ?

drjova: No, we are anyway removing them https://github.com/SCOAP3/scoap3-next/blob/master/scoap3/modules/workflows/workflows/articles_upload.py#L473 I would say to ignore empty and None values

Neodia: Added check and test 👍

drjova (approved):
LGTM, just the fix for the ref in the commit message.

PR #19, opened 2022-05-13
Generic Parsing: Added fix_publication_date

Neodia:
* Ref: cern-sis/issues-scoap3#52

### DIFF
--- a/dags/common/parsing/generic_parsing.py
+++ b/dags/common/parsing/generic_parsing.py
@@ -1,3 +1,6 @@
+from datetime import date
+
+
 def take_first(arr):
     try:
         return next(filter(None, arr))
@@ -19,3 +22,15 @@ def classification_numbers(arr, standard="PACS"):
 
 def free_keywords(arr, source="author"):
     return [{"source": source, "value": val} for val in arr if val]
+
+
+def fix_publication_date(pub_date):
+    if not pub_date:
+        return ""
+
+    def inner_fix(splitted):
+        return splitted if len(splitted) == 3 else inner_fix(splitted + ["1"])
+
+    year, month, day = map(int, inner_fix(pub_date.split("-")))
+    tmp_date = date(year, month, day)
+    return f"{tmp_date.year:04d}-{tmp_date.month:02d}-{tmp_date.day:02d}"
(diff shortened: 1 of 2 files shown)


dags/common/parsing/generic_parsing.py
```diff
@@ -19,3 +22,19 @@ def classification_numbers(arr, standard="PACS"):
...
+
+
+def fix_publication_date(pub_date):
+    if not pub_date:
+        return ""
+
+    def inner_fix(splitted):
+        return splitted if len(splitted) == 3 else inner_fix(splitted + ["1"])
+
+    splitted_date = inner_fix(pub_date.split("-"))
```
drjova:
```suggestion
    year, month, day = map(int, inner_fix(pub_date.split("-")))
```

dags/common/parsing/generic_parsing.py
```diff
@@ -19,3 +22,19 @@ def classification_numbers(arr, standard="PACS"):
...
+    tmp_date = date(
+        int(splitted_date[0]),
+        int(splitted_date[1]),
+        int(splitted_date[2]),
+    )
+    return f"{tmp_date.year:04d}-{tmp_date.month:02d}-{tmp_date.day:02d}"
```
drjova, line 36: Why not just `tmp_date.strftime('%Y-%m-%d')`?

Neodia: Because string formats are not the same in MacOS/Linux. Setting the date like this normalizes the format in all OS

dags/common/parsing/generic_parsing.py
Neodia: That's great, updated 👍

dags/common/parsing/generic_parsing.py
```diff
@@ -19,3 +22,19 @@ def classification_numbers(arr, standard="PACS"):
...
+        return ""
+
+    def inner_fix(splitted):
+        return splitted if len(splitted) == 3 else inner_fix(splitted + ["1"])
+
+    year, month, day = map(int, inner_fix(pub_date.split("-")))
+    tmp_date = date(
+        int(year),
+        int(month),
+        int(day),
```
drjova: you don't need the `int` :)

Neodia: Brainfart sorry, fixed 👍

drjova (approved):

PR #20, opened 2022-05-13
Cleanup: remove not save attributes from tags

ErnestaP:
* Added added cleaning html from some attributes and tags
* Added tests
* ref: cern-sis/issues-scoap3#38

### DIFF
--- a/dags/common/cleanup.py
+++ b/dags/common/cleanup.py
@@ -1,11 +1,23 @@
 def clean_whitespace_characters(input):
     return " ".join(input.split())
 
+
 def convert_html_subsripts_to_latex(input):
     from re import sub
 
     input = sub("<sub>(.*?)</sub>", r"$_{\1}$", input)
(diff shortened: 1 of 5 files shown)


dags/common/cleanup.py
```diff
@@ -1,11 +1,23 @@
...
+
+def clean_html(input):
+    from lxml.etree import ParserError, XMLSyntaxError
+    from lxml.html.clean import Cleaner
+
+    try:
+        cleaner = Cleaner(safe_attrs_only=True, remove_unknown_tags=False)
+        return cleaner.clean_html(input)
+    except (XMLSyntaxError, ParserError):
+        return input
```
drjova, line 23: same with https://github.com/cern-sis/workflows/pull/21 [bleach](https://github.com/mozilla/bleach) :)

ErnestaP:
Tags together with their attributes are removed on https://github.com/cern-sis/issues-scoap3/issues/36. We don't need to do it again

PR #21, opened 2022-05-13
Cleanup: remove specific tags

ErnestaP:
* Added added cleaning html from specific tags
* Added tests
* ref: cern-sis/issues-scoap3#36


dags/common/cleanup.py
```diff
@@ -1,11 +1,18 @@
...
     input = sub("<sub>(.*?)</sub>", r"$_{\1}$", input)
     input = sub("<inf>(.*?)</inf>", r"$_{\1}$", input)
-    input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)   
+    input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)
 
     return input
+
+
+def remove_specific_tags(value, which_ones=(), keep=()):
+    from w3lib.html import remove_tags
```
drjova: it's better to use [bleach](https://github.com/mozilla/bleach) :)

dags/common/cleanup.py
```diff
@@ -1,11 +1,19 @@
...
+    input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)
 
     return input
+
+
+def remove_specific_tags(value, keep=[]):
+    from bleach.sanitizer import Cleaner
+
+    cleaner = Cleaner(tags=keep, strip=True)
+    return cleaner.clean(value).replace("\n", "")
```
drjova:
```suggestion
    from bleach import clean
    return clean(value, strip=True, tags=keep)
```

Do we need to remove the new lines? Was it the case before?

tests/units/common/test_cleanup.py
```diff
@@ -27,36 +32,97 @@ def test_clean_whitespace_characters(test_input, expected):
...
+        pytest.param(xml, xml_div_and_h1, ["div", "h1"], id="test_keep_div_and_h1"),
+    ],
+)
+def test_remove_specific_tags(test_input, expected, params):
+    output = remove_specific_tags(test_input, keep=params)
+    assert output == expected
+
+
+def test_test_remove_specific_tags_with_mathML(shared_datadir):
+    import os
```
drjova: the imports should be in the beginning of the module 😃

drjova: `shared_datadir` is from pytest-datadir right? you only need the following:

```python
def test_read_global(shared_datadir):
    contents = (shared_datadir / 'hello.txt').read_text()
    assert contents == 'Hello World!\n'
```

dags/common/cleanup.py
ErnestaP: yes, because strip adds new lines, instead if the tag, which was inside of the allowed tag.
For example, a is allowed:
input:
``` <a><b>'example'</b></a>```
output: 
```
<a>
'example'
</a>
```
No, it wasn't a case before

Comment (drjova) [reply dags/common/cleanup.py]: @ErnestaP thank you so much for the explanation! Testing this case it doesn't seem to be the case:

```python
In [1]: from bleach import clean                                                                                                                                                                                                                                                                                                                                   
In [2]: test = "<a><b>'example'</b></a>"                                                                                                                                                                                                                                                                                                                           
In [3]: clean(test, strip=True, tags=['a'])                                                                                                                                                                                                                                                                                                                        
Out[3]: "<a>'example'</a>"
```
Additionally, you ignored my suggestion :) You don't need to initialise the class you just use `clean(value, strip=True, tags=keep)` as in [documentation](https://bleach.readthedocs.io/en/latest/#basic-use)

ErnestaP: sorry for ignoring the suggestion.
I see that I made a mistake in the example, it has to be without quotes:
```<a><b>example</b></a>```

Comment (ErnestaP) [reply dags/common/cleanup.py]:
```
In [4]: xml = "<div><p>example<h1> h1 example</h1></p></div>"
In [5]: clean(xml, strip=True, tags=['p'])
Out[5]: '<p>example\n h1 example</p>'
```

File: dags/common/cleanup.py [hunk 4]
```diff
@@ -1,11 +1,19 @@
...
 
     input = sub("<sub>(.*?)</sub>", r"$_{\1}$", input)
     input = sub("<inf>(.*?)</inf>", r"$_{\1}$", input)
-    input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)   
+    input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)
 
     return input
+
+
+def remove_specific_tags(value, tags=[], attributes=[]):
```
drjova: Just for references why mutable arguments in function's signature are a really bad practise (I know you just copy-paste it) :) https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/   

I would suggest the following change:

```suggestion
def remove_specific_tags(value, tags=None, attributes=None):
    if tags is None:
        tags = []
    if attributes is None:
        attributes = []
```

drjova:
@ErnestaP could you please install bleach 4.1.0 instead of 5.0.0? They recently introduce this change. Thank you for checking ☺️

requirements.txt
```diff
@@ -11,3 +11,4 @@ pytest==7.1.0
 PyYAML==6.0
 furl==2.1.3
 structlog==21.5.0
+bleach==4.0.0
```
drjova: why 4.0.0 and not 4.1.0 as suggested in the [comment](https://github.com/cern-sis/workflows/pull/21#issuecomment-1127902016) 😅

dags/common/cleanup.py
```diff
@@ -1,11 +1,23 @@
...
+    input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)
 
     return input
+
+
+def remove_specific_tags(value, tags=None, attributes=None):
+    if tags is None:
+        tags = []
+    if attributes is None:
+        attributes = []
```
Neodia:
```suggestion
     tags = tags or []
     attributes = attributes or []
```

tests/units/common/test_cleanup.py
```diff
@@ -34,29 +39,98 @@ def test_clean_whitespace_characters(test_input, expected):
...
+        ),
+    ],
+)
+def test_remove_specific_tags(test_input, expected, tags, attributes):
+    assert (
+        remove_specific_tags(test_input, tags=tags, attributes=attributes) == expected
+    )
+
 
+def test_test_remove_specific_tags_with_mathML(shared_datadir):
```
Neodia: `test_` prefix written two times

Neodia (approved):

PR #24, opened 2022-05-16
Generic Parsing: Added publication_info

Neodia:
* Ref: cern-sis/issues-scoap3#54

### DIFF
--- a/dags/common/parsing/generic_parsing.py
+++ b/dags/common/parsing/generic_parsing.py
@@ -77,3 +77,19 @@ def parse_thesis_supervisors(value):
         "full_name": value.get("full_name"),
         "affiliation": value.get("affiliation"),
     }
+
+
+def publication_info(article):
+    return [
+        {
+            "journal_title": article.pop("journal_title", ""),
+            "journal_volume": article.pop("journal_volume", ""),
+            "year": int(article.pop("journal_year", 0)) or "",
+            "journal_issue": article.pop("journal_issue", ""),
+            "artid": article.pop("journal_artid", ""),
+            "page_start": article.pop("journal_fpage", ""),
+            "page_end": article.pop("journal_lpage", ""),
+            "material": article.pop("journal_doctype", ""),
+            "pubinfo_freetext": article.pop("pubinfo_freetext", ""),
+        }
+    ]
--- a/tests/units/common/parsing/test_generic_parsing.py
+++ b/tests/units/common/parsing/test_generic_parsing.py
@@ -7,6 +7,7 @@ from common.parsing.generic_parsing import (
     list_to_value_dict,
     parse_authors,
     parse_thesis_supervisors,
+    publication_info,
     split_fullname,
     take_first,
 )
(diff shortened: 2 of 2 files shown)


dags/common/parsing/generic_parsing.py
```diff
@@ -77,3 +77,19 @@ def parse_thesis_supervisors(value):
...
         "affiliation": value.get("affiliation"),
     }
+
+
+def publication_info(article):
+    return [
+        {
+            "journal_title": article.pop("journal_title", ""),
+            "journal_volume": article.pop("journal_volume", ""),
+            "year": int(article.pop("journal_year", 0)) or "",
```
ErnestaP, line 87: Could you please explain why we need an empty string here (with or)?
Because you'll get 0 if the value journal_year is not in dict, and if it's 0 OR will return an empty string.  
Is there a specific purpose for it?

tests/units/common/parsing/test_generic_parsing.py
```diff
@@ -189,3 +190,31 @@ def test_parse_authors(test_input, expected):
...
+            "journal_volume": "Test Value",
+            "year": 2022,
+            "journal_issue": "Test Value",
+            "artid": "",
+            "page_start": "",
+            "page_end": "",
+            "material": "",
+            "pubinfo_freetext": "",
+        }
+    ] == publication_info(test_article)
```
ErnestaP: I think, would be nice to add more tests, for example, if some of the fields are not in the dict. What do you think?
To cover 'not usual/expected cases as well

tests/units/common/parsing/test_generic_parsing.py
```diff
@@ -189,3 +190,31 @@ def test_parse_authors(test_input, expected):
...
     assert expected == parse_thesis_supervisors(test_input)
+
+
+def test_publication_info():
+    test_article = {
+        "journal_title": "Test Value",
+        "journal_volume": "Test Value",
+        "journal_year": "2022",
+        "journal_issue": "Test Value",
+        "journal_artid": "",
```
drjova: could you please add a test for the happy path as well (all the values are populated)?

Neodia: Done

tests/units/common/parsing/test_generic_parsing.py
Neodia: Yes good point, added

dags/common/parsing/generic_parsing.py
Neodia: No I cannot say for sure. My guess is data representation : the year 0 is a valid year, whereas "" defines something that is empty. @drjova do you have some insight ?

ErnestaP: I understand that 0 will never be a value in this case: 
int(article.pop("journal_year", 0)) or ""
and
int(article.pop("journal_year", "")) 
give the same result, do you agree?

ErnestaP: My only doubt is this one. 
If it will be an empty string, might cause a problem, because it might be int according to [schema](https://repo.scoap3.org/schemas/hep.json), am I right?

drjova: @Neodia this is wrong unfortunately, `year` should be an int:
```
{
  maximum: 2050,
  minimum: 1000,
  type: integer
}
```

PR #25, opened 2022-05-16
Cleanup: clean collaboration

ErnestaP:
* Removed "for the" from collaboration string
* Added tests
* ref: cern-sis/issues-scoap3#39


dags/common/cleanup.py
```diff
@@ -1,11 +1,17 @@
...
     input = sub("<sub>(.*?)</sub>", r"$_{\1}$", input)
     input = sub("<inf>(.*?)</inf>", r"$_{\1}$", input)
-    input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)   
+    input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)
 
     return input
+
+
+def clean_collaboration(input):
+    """Remove the prefixes for collaborations"""
```
Neodia: We dont need this comment

tests/units/common/test_cleanup.py
```diff
@@ -27,36 +31,99 @@ def test_clean_whitespace_characters(test_input, expected):
...
+        pytest.param(
+            collaboration_string_cleaned,
+            collaboration_string_cleaned,
+            id="test_collaboration_string_with_cleaned_value",
+        ),
+    ],
+)
+def test_clean_collaboration(test_input, expected):
+    output = clean_collaboration(test_input)
+    assert output == expected
```
Neodia: No need to store in output var

dags/common/cleanup.py
ErnestaP: what is wrong with this comment?

Neodia: What is wrong : It's a comment.
Let me explain: There are few cases there commenting makes sense imo. Code should be self explanatory. (Here it clearly is, we can clearly see that what you're doing is : clean whitespace characters and replace "for the".)
If your code is not self explanatory, maybe some named functions should be used instead to make it more readable.
Comments might be useful when implementing complex algorithms maybe. We could still argue that we could split the algo in different smaller functions to make it more understandable.

This is all my opinion. I'm open to discussion. Any arguments against it @ErnestaP @drjova ?

Neodia: @ErnestaP This was stuck as a pending review, sorry for the late reply

Neodia (approved):

dags/common/cleanup.py
ErnestaP: I see your point. I agree that code has to be self-explanatory and this is a docstring, which just explains what functions do. For me, because I don't know what collaboration string looks like, gives me a bit of sense of why we have to remove these two words.
I would add docstring on almost all functions. That is what I was taught. Of course, this docstring which is in this PR is not the best example.

PR #26, opened 2022-05-17
Generic Parsing: Added merge_dois

Neodia:
* Modified publication_info to prevent side effects
* Added clear_unnecessary_fields to clear the article afterwards
* Ref: cern-sis/issues-scoap3#58

### DIFF
--- a/dags/common/parsing/generic_parsing.py
+++ b/dags/common/parsing/generic_parsing.py
@@ -82,14 +82,36 @@ def parse_thesis_supervisors(value):
 def publication_info(article):
     return [
         {
-            "journal_title": article.pop("journal_title", ""),
-            "journal_volume": article.pop("journal_volume", ""),
-            "year": int(article.pop("journal_year", 0)) or "",
-            "journal_issue": article.pop("journal_issue", ""),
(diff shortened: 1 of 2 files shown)


ErnestaP (approved):

dags/common/parsing/generic_parsing.py
```diff
@@ -82,14 +82,36 @@ def parse_thesis_supervisors(value):
...
+            "page_start": article.get("journal_fpage", ""),
+            "page_end": article.get("journal_lpage", ""),
+            "material": article.get("journal_doctype", ""),
+            "pubinfo_freetext": article.get("pubinfo_freetext", ""),
         }
     ]
+
+
+def merge_dois(article):
+    return article["dois"] + article.get("related_article_doi", [])
```
drjova, line 99: are we sure that the `dois` is a list?

Neodia: Dois are being handled like this in Hepcrawl :
`dois_out = ListToValueDict()`
So I think so yes.

For the Springer parser I'm encapsulating the value in an array anyways

PR #27, opened 2022-05-17
Cleanup: clean affiliations value for authors

ErnestaP:
* Removed "label" tag and its content, removed all tags and white spaces from affiliation value
* Added tests
* ref: cern-sis/issues-scoap3#37

### DIFF
--- a/dags/common/cleanup.py
+++ b/dags/common/cleanup.py
@@ -1,4 +1,6 @@
 from bleach import clean
+from bleach.html5lib_shim import Filter
+from bleach.sanitizer import Cleaner
 
 
 def clean_whitespace_characters(input):
@@ -11,8 +13,10 @@ def convert_html_subsripts_to_latex(input):
     input = sub("<sub>(.*?)</sub>", r"$_{\1}$", input)
     input = sub("<inf>(.*?)</inf>", r"$_{\1}$", input)
     input = sub("<sup>(.*?)</sup>", r"$^{\1}$", input)
-
-    return input
+    return input 
+    
+def clean_collaboration(input):
+    return clean_whitespace_characters(input.replace("for the", ""))
 
 
 def clean_collaboration(input):
@@ -22,3 +26,31 @@ def remove_specific_tags(value, tags=None, attributes=None):
     tags = tags or []
     attributes = attributes or []
     return clean(value, tags=tags, attributes=attributes, strip=True)
+
+
+class RemoveLabelTagsContentFilter(Filter):
+    def __iter__(self):
+        label_tag = False
+        for token in super().__iter__():
+            if token["type"] == "StartTag" and token["name"] == "label":
+                label_tag = True
+            elif token["type"] == "EndTag" and token["name"] == "label":
+                label_tag = False
+            elif label_tag:
+                token["data"] = ""
+            yield token
+
+
+def clean_affiliation_for_author(input):
+    cleaner = Cleaner(
+        filters=[RemoveLabelTagsContentFilter], tags=["label"], strip=True
+    )
+    cleaned_label_content = cleaner.clean(input)
+    return clean_whitespace_characters(remove_specific_tags(cleaned_label_content))
+
+
+
+def clean_all_affiliations_for_author(input):
+    for affiliation in input.get("affiliations", []):
+        affiliation["value"] = clean_affiliation_for_author(affiliation["value"])
+    return input
--- a/tests/units/common/test_cleanup.py
+++ b/tests/units/common/test_cleanup.py
@@ -1,5 +1,7 @@
 import pytest
 from common.cleanup import (
+    clean_affiliation_for_author,
+    clean_all_affiliations_for_author,
     clean_collaboration,
     clean_whitespace_characters,
     convert_html_subsripts_to_latex,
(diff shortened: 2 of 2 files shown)


dags/common/cleanup.py
```diff
@@ -18,7 +20,37 @@ def convert_html_subsripts_to_latex(input):
...
+            elif token["type"] == "EndTag":
+                label_tag = False
+            elif label_tag:
+                token["data"] = ""
+            yield token
+
+
+def clean_affiliation_for_author(input):
+    cleaner = Cleaner(
+        filters=[RemoveUnneccessaryTagsFilter], tags=["label"], strip=True
```
drjova: @ErnestaP I really like your solution with the filter 👏

dags/common/cleanup.py
```diff
@@ -18,7 +20,37 @@ def convert_html_subsripts_to_latex(input):
...
 
+
 def remove_specific_tags(value, tags=None, attributes=None):
     tags = tags or []
     attributes = attributes or []
     return clean(value, tags=tags, attributes=attributes, strip=True)
+
+
+class RemoveUnneccessaryTagsFilter(Filter):
+    """Filter removes content of label, but not the tag itself!"""
```
drjova: there is a python guide [PEP257](https://peps.python.org/pep-0257/) how to write python doc strings 😁  Please have a look. In general, the function/class name should be enough to understand what it does and we usually omit unnecessary docs/comments 😃 Hence, I would suggest to remove this comment.

dags/common/cleanup.py
```diff
@@ -18,7 +20,37 @@ def convert_html_subsripts_to_latex(input):
...
+
+    def __iter__(self):
+        label_tag = False
+        for token in Filter.__iter__(self):
+            if token["type"] == "StartTag" and token["name"] == "label":
+                label_tag = True
+            elif token["type"] == "EndTag":
+                label_tag = False
+            elif label_tag:
+                token["data"] = ""
```
drjova, line 41: shouldn't be something like:

```suggestion
            if label_tag:
                token["data"] = ""
                label_tag = False
```

dags/common/cleanup.py
```diff
@@ -18,7 +20,37 @@ def convert_html_subsripts_to_latex(input):
...
     return clean_whitespace_characters(input.replace("for the", ""))
 
+
 def remove_specific_tags(value, tags=None, attributes=None):
     tags = tags or []
     attributes = attributes or []
     return clean(value, tags=tags, attributes=attributes, strip=True)
+
+
+class RemoveUnneccessaryTagsFilter(Filter):
```
drjova: because the class is targetting specifically `label` WDYT to rename it to something like: `RemoveLabelTagsFilter`?

dags/common/cleanup.py
drjova: shouldn't be something like:
```suggestion
            if token["type"] == "StartTag" and token["name"] == "label":
                 token["data"] = ""
```

tests/units/common/test_cleanup.py
```diff
@@ -169,3 +182,69 @@ def test_remove_specific_tags_with_mathML(shared_datadir):
...
+    [
+        pytest.param(
+            test_object,
+            exptected_test_object,
+            id="test_clean_affiliation_for_author",
+        ),
+        pytest.param(
+            empty,
+            empty,
+            id="test_clean_affiliation_for_author",
```
Neodia: The ids are the same

dags/common/cleanup.py
ErnestaP: We cannot use it like this, because the output regarding the label of iteration is:
```
....
{'type': 'StartTag', 'name': 'label', 'namespace': None, 'data': {}}
{'data': 'Label', 'type': 'Characters'}
{'type': 'EndTag', 'name': 'label', 'namespace': None}
....
```
We need to change the value in 2nd line ;)

drjova: Ok I see, thanks for the explanation, we should change it a bit just to make it more explicit:

```

 token["type"] == "EndTag" and token["name"] == "label":
     label_tag = False


     
```

dags/common/cleanup.py
```diff
@@ -18,7 +20,35 @@ def convert_html_subsripts_to_latex(input):
...
 def remove_specific_tags(value, tags=None, attributes=None):
     tags = tags or []
     attributes = attributes or []
     return clean(value, tags=tags, attributes=attributes, strip=True)
+
+
+class RemoveLabelTagsContentFilter(Filter):
+    def __iter__(self):
+        label_tag = False
+        for token in Filter.__iter__(self):
```
Neodia: Can't this be turned to `super().__iter__()` ?

ErnestaP: Good question! 
In line 32 we overwriting RemoveLabelTagsContentFilter `__iter__` function (which is built-in func.)  adding label attribute. Then in line 33 ` Filter.__iter__(self)` we iterating per this overwritten func. in RemoveLabelTagsContentFilter class. 
If we will do `super().__iter__()`, we will pass not self of our new class but self of parent class - so in the end we will overwrite the parent class `__iter__` rather than RemoveLabelTagsContentFilter. Is it clear?

drjova: Good point @Neodia  and thanks for the explanation @ErnestaP. There is no need to pass `self` when you call a class method for example:


```python

class Something:

    def hello(self):
        print('hello')
       
Something().hello() # <---- we are not passing `self`
```

drjova: Could you please update it with `super()`?

Neodia (approved):

PR #28, opened 2022-05-19
Generic Parsing: Created generic parsing for all publishers

Neodia:
* Modified parser to have a generic function
* Parser's parse first parses publisher rules, then generic
* Springer: Abstract now handles TEX expressions
* Modified tests
* Ref: cern-sis/issues-scoap3#61


dags/common/parsing/parser.py
```diff
@@ -10,13 +29,63 @@ class IParser:
...
+    def _generic_parsing(self, publisher_parsed_article: Dict):
+        def get(field, default=[]):
+            return publisher_parsed_article.get(field, default)
+
+        parsed_article = publisher_parsed_article.copy()
+
+        parsed_article.update(
+            {
+                "authors": [parse_author(author) for author in get("authors", [])],
+                "abstract": str.strip(get("abstract", "")),
```
ErnestaP: I see you already doing clean here. Maybe then is better to remove all repeated whitespace chars, not just trailing ones?

dags/common/parsing/parser.py
```diff
@@ -10,13 +29,63 @@ class IParser:
...
+                "collaborations": list_to_value_dict(get("collaborations", "")),
+                "collections": list_to_value_dict(get("collections"), "primary"),
+                "title": str.strip(get("title")),
```
ErnestaP: same here

dags/common/parsing/parser.py
```diff
@@ -10,13 +29,63 @@ class IParser:
...
+                "subtitle": pipe_functions(
+                    [lambda x: [str.strip(subtitle) for subtitle in x], take_first],
```
ErnestaP: and here 😄

dags/common/parsing/parser.py
```diff
@@ -10,13 +29,63 @@ class IParser:
...
+                    get("subtitle"),
+                ),
+                "journal_year": get("journal_year"),
+                "preprint_date": take_first(get("preprint_date")),
+                "date_published": fix_publication_date(get("date_published")),
+                "related_article_doi": list_to_value_dict(get("related_article_doi")),
+                "free_keywords": free_keywords(get("free_keywords")),
```
ErnestaP: if we're doing cleaning from whitespace chars already in almost all fields in the parser, I would suggest doing it in all the rest fields as well: free_keywords and affiliations. ☺️

tests/units/springer/test_parser.py
```diff
@@ -37,7 +37,7 @@ def extract_zip_to_article(zip_filename):
 
 @fixture()
 def parsed_articles(parser, articles):
-    return [parser.parse(article) for article in articles]
+    return [parser._publisher_specific_parsing(article) for article in articles]
```
ErnestaP, line 39: NOTE: this comment for line 14 (cannot put a comment there), is it possible to use https://pypi.org/project/pytest-datadir/?

dags/common/parsing/parser.py
Neodia: This or I'll do it for no field and then we'll do it during cleaning. WDYT ?

tests/units/springer/test_parser.py
Neodia: This will duplicate data as we using this dir for the `docker-compose.yml` FTP service. I think it's alright if we let it as is

dags/common/parsing/parser.py
Neodia: Done

dags/common/parsing/parser.py
Neodia: Done

dags/common/parsing/parser.py
Neodia: Done

dags/common/parsing/parser.py
Neodia: Done

ErnestaP (approved):

dags/common/parsing/generic_parsing.py
```diff
@@ -115,3 +115,7 @@ def clear_unnecessary_fields(article):
     ]
     [new_article.pop(field_name, "") for field_name in field_list]
     return new_article
+
+
+def remove_none_entires(article):
+    return {key: val for (key, val) in article.items() if val is not None}
```
drjova: 👍  Let's do this at the end of the parsing, including nested elements and removing any empty values including `""`. A proposal would be something like [that](https://github.com/inspirehep/inspire-dojson/blob/640fc41e05261b1e40005d916740f3c67bdbd8f1/inspire_dojson/utils/__init__.py#L179-L198)

tests/units/common/parsing/test_parser.py
```diff
@@ -20,13 +20,142 @@
     </Root>
 """
 
+publisher_parsed_article = {
```
drjova: @Neodia I would move these to a separate file (fixture) and use https://pypi.org/project/pytest-datadir/ to read it (we already have it)

tests/units/springer/test_parser.py
drjova: good point, we should have separate data for tests even if they are duplicates.

dags/springer/parser.py
```diff
@@ -110,6 +112,15 @@ def __init__(self) -> None:
         super().__init__(extractors)
 
     def _get_abstract(self, article: ET.Element):
+        def is_latex(text):
+            return (
+                re.search(r"\\\[(.+?)\\\]", text)
+                or re.search(r"\\\((.+?)\\\)", text)
+                or re.search(r"\\\[(.+?)\\\]", text)
+                or re.search(r"\\begin\{displaymath\}(.+?)\\end\{displaymath\}", text)
+                or re.search(r"\$\$(.+)\$\$", text)
```
drjova: latex not always are between `$` can we just get the value from `CDATA`?

dags/springer/parser.py
```diff
@@ -118,13 +129,15 @@ def _get_abstract(self, article: ET.Element):
             [child_text for child_text in child.itertext()] for child in paragraph
         ]
         text_to_skip_flatten = [
-            item for sublist in text_to_skip_arrays for item in sublist
+            item
+            for sublist in text_to_skip_arrays
+            for item in sublist
+            if not is_latex(item)
```
drjova: I just realize this code, we can make it more readable with recursion for an example can be:

```python
def flatten_list(input_list):
      return [
          element for innerList in input_list for element in flatten_list(innerList)
      ]
```

tests/units/common/parsing/test_parser.py
Neodia: Ernesta also gave the idea of `pytest-datadir`. I said it was a bad idea because it would copy data as I'm using the same that we are pushing to the `docker-compose` FTP server. What do you think ?

Neodia: Just saw your other comment. Will change this

dags/springer/parser.py
Neodia: I don't totally get the suggestion. Recursion needs a ending condition. Here there is none and we will be trying to iterate through a int and it will fail. Am I missing something ?

dags/springer/parser.py
Neodia: No because the LaTeX expressions are not always in CDATA. They seems to be between tags with `format="TEX"`. Anyways as discussed, I need to change this to remove the regex

dags/springer/parser.py
drjova: my bad I didn't completely finish my suggestion:

```python
def flatten_list(input_list):
    if isinstance(input_list, (list, tuple)):
        return [
            element for innerList in input_list for element in flatten_list(innerList)
        ]
    return [input_list]
```

Neodia: I find that your code is the same readability than mine (because we still have 2 `for`). Yours is good to handle lists that have more than two levels which is not the case here. Do you think that I should still change it ?

drjova: I was having having trouble to understand why we need too many loops (including `text_to_skip_arrays`), maybe i'm missing something, but can we just reduce them by doing the following:

```python
# no need of `text_to_skip_arrays` and 
text_to_skip_flatten = [
     child_text 
     for child in paragraph 
     for child_text in child.itertext() 
     if not is_latex(child_text)
]
```

One more question instead of `child.itertext` can we just get `chid.text` why we need to iterate? this will reduce the loops to 1:

```python
text_to_skip_flatten = [
     text 
     for text in paragraph
     if not is_latex(text)
]

# or 
filter(lambda x: not is_latex(x), paragraph)
```

WDYT?

Neodia: Reasoning for `text_to_skip_arrays`: All children do not have valuable data. We need to first identify the data that we don't need to skip it.

Reasoning for `child.itertext`: `child.text` only returns the text from the current node. `itertext` returns from current node and from children, and children inside it.

I will check again this function anyways to see if it cannot be rewritten in some way

tests/units/springer/test_parser.py
Neodia: Done

dags/common/parsing/generic_parsing.py
Neodia: Done

tests/units/common/parsing/test_parser.py
Neodia: Done

dags/springer/parser.py
Neodia: Latex expressions are now coming from `<EquationSource format="TEXT>` tags 👍

dags/springer/parser.py
Neodia: Sorry I didn't get your point on text to skip before. Changed, let me know if it's better 👍

tests/units/common/parsing/test_parser.py
drjova: @Neodia why not moving these two in a separate `json` files?

dags/springer/parser.py
```diff
@@ -110,21 +112,23 @@ def __init__(self) -> None:
         super().__init__(extractors)
 
     def _get_abstract(self, article: ET.Element):
+        def is_latex_node(node: ET.Element):
+            return node.tag == "EquationSource" and node.attrib["Format"] == "TEX"
```
drjova, line 116: would this take also the `<CDATA` text?

drjova (approved):
LGTM just the two comments!

dags/springer/parser.py
Neodia: As long as the CDATA is in the `<EquationSource format="TEXT>` tag yes. (It seems to be there)

tests/units/common/parsing/test_parser.py
Neodia: Sorry I mingled this comment with the other with the datadir. Done 👍

PR #29, opened 2022-05-20
Enhance steps

ErnestaP:


dags/common/enhance.py
```diff
@@ -0,0 +1,54 @@
+import datetime
+
+
+class Enhancement:
```
Neodia: I would rename this to `Enhancer`.

dags/common/enhance.py
```diff
@@ -0,0 +1,54 @@
...
+    def construct_titles(self, item, publisher):
+        item["titles"] = [
+            {
+                "title": item.pop("title", ""),
+                "subtitle": item.pop("subtitle", ""),
+                "source": publisher,
+            }
+        ]
+
+    def __call__(self, publisher, item):
```
Neodia: I think I would prefer this function to return a copy of the object, so we keep the state of the input object

Neodia: I'm not a fan of using the __call__ function to enhance. I would define a function called `enhance`

dags/common/enhance.py
```diff
@@ -0,0 +1,54 @@
...
+            "source": publisher,
+            "method": publisher,
+            "date": self.creation_date,
+            "submission_number": "path/to/the/file",
+        }
+
+    def construct_copyright(self, item):
+        item["copyright"] = [
+            {
+                "holder": item.pop("copyright_holder", ""),
```
Neodia: To prevent side effects in the input item, I would change `pop` to `get`.

dags/common/enhance.py
```diff
@@ -0,0 +1,54 @@
+import datetime
+
+
+class Enhancement:
+    def __init__(self):
+        self.creation_date = datetime.datetime.now().isoformat()
+
+    def constuct_abstracts(self, item, publisher):
+        item["abstracts"] = [{"value": item.pop("abstracts", ""), "source": publisher}]
```
Neodia: To prevent side effects in the input item, I would change `pop` to `get`.

dags/common/enhance.py
```diff
@@ -0,0 +1,54 @@
+import datetime
+
+
+class Enhancement:
+    def __init__(self):
+        self.creation_date = datetime.datetime.now().isoformat()
```
Neodia: Creation date should not be a property of the enhancer IMO. We should declare this at the start of the __call__ function and pass it to the functions when needed

dags/common/enhance.py
```diff
@@ -0,0 +1,54 @@
...
+            {"date": item.pop("date_published", ""), "publisher": publisher}
+        ]
+
+    def construct_record_creation_date(self, item):
+        item["record_creation_date"] = self.creation_date
+
+    def construct_titles(self, item, publisher):
+        item["titles"] = [
+            {
+                "title": item.pop("title", ""),
```
Neodia: Same for the `pop` here

tests/units/common/test_enhance.py
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+
+@pytest.fixture(scope="function")
+def geneirc_pseudo_parser_output():
+    yield PseudoParsesOutput()
+
+
+publisher = "Springer"
+expexted_output = {
```
Neodia: typo

Neodia:
I like how the Enhancer class turned out. It's simple, concise and clear 👍

dags/common/enhance.py
```diff
@@ -0,0 +1,54 @@
+import datetime
+
+
+class Enhancement:
+    def __init__(self):
+        self.creation_date = datetime.datetime.now().isoformat()
+
+    def constuct_abstracts(self, item, publisher):
```
Neodia: These functions are only used in this class, thus we should add a leading underscore in the name, as described in [PEP 8](https://peps.python.org/pep-0008/)

> Use one leading underscore only for non-public methods and instance variables.

dags/common/enhance.py
ErnestaP: good idea

dags/common/enhance.py
ErnestaP: Good proposal. For me **call** fits better here, since the class does just one job- enhance a given dict.  There cannot be any more actions done with that

dags/common/enhancer.py
```diff
@@ -0,0 +1,52 @@
...
+        item["titles"] = [
+            {
+                "title": item.pop("title", ""),
+                "subtitle": item.pop("subtitle", ""),
+                "source": publisher,
+            }
+        ]
+
+    def __call__(self, publisher, item):
+        self.creation_date = datetime.datetime.now().isoformat()
```
Neodia: You are still defining the creation date as a class property

ErnestaP: 🤦‍♀️

dags/common/enhance.py
Neodia: Yes I get your point. I still don't really like it but it's in domain of personal preference here so nothing to change 👍

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,97 @@
...
+            "copyright_statement": "copyright_statement",
+            "copyright_material": "copyright_material",
+            "date_published": "2022-05-20",
+            "title": "title",
+            "subtitle": "subtitle",
+        }
+
+
+@pytest.fixture(scope="function")
+def geneirc_pseudo_parser_output():
```
Neodia: typo

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,97 @@
...
+    yield PseudoParsesOutput()
```
Neodia: This should directly return the value instead. `PseudoParsesOutput` class is not needed. Just return the dict present in PseudoParsesOutput.get_ouput directly here

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,97 @@
...
+@freeze_time("2022-05-20")
+def test_all_contructors(geneirc_pseudo_parser_output):
+    output = geneirc_pseudo_parser_output
+    generic_parser_output = output.get_output()
+    enhancement = Enhancer()
+    enhanced = enhancement(item=generic_parser_output, publisher=publisher)
+    assert enhanced == expected_output
+
+
+class EmptyPseudoParsesOutput:
```
Neodia: Here same than for `PseudoParsesOutput `

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,99 @@
+import pytest
+from common.enhancer import Enhancer
+from freezegun import freeze_time
+
+
+class PseudoParsesOutput:
```
Neodia: This seems to be unused

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,99 @@
...
+    ],
+    "imprints": [{"date": "2022-05-20", "publisher": publisher}],
+    "record_creation_date": "2022-05-20T00:00:00",
+    "titles": [{"title": "title", "subtitle": "subtitle", "source": publisher}],
+}
+
+
+@freeze_time("2022-05-20")
+def test_all_contructors(generic_pseudo_parser_output):
+    output = generic_pseudo_parser_output
```
Neodia: Output var not neede

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,99 @@
...
+        "copyright_statement": "copyright_statement",
+        "copyright_material": "copyright_material",
+        "date_published": "2022-05-20",
+        "title": "title",
+        "subtitle": "subtitle",
+    }
+
+
+publisher = "Springer"
+expected_output = {
```
Neodia: I think it's better if we put those as fixtures

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,99 @@
...
+        "copyright_year": "",
+        "copyright_statement": "",
+        "copyright_material": "",
+        "date_published": "",
+        "title": "",
+        "subtitle": "",
+    }
+
+
+expected_output_from_empty_input = {
```
Neodia, line 52: Same here. Put it as fixture or directly in the function as it seems to be used only once

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,99 @@
...
+            "copyright_year": "2020",
+            "copyright_statement": "copyright_statement",
+            "copyright_material": "copyright_material",
+            "date_published": "2022-05-20",
+            "title": "title",
+            "subtitle": "subtitle",
+        }
+
+
+@pytest.fixture(scope="function")
```
Neodia: `function` scope is not needed as it is the default behaviour
https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes

tests/units/common/test_enhancer.py
```diff
@@ -0,0 +1,89 @@
...
+        "copyright_year": "2020",
+        "copyright_statement": "copyright_statement",
+        "copyright_material": "copyright_material",
+        "date_published": "2022-05-20",
+        "title": "title",
+        "subtitle": "subtitle",
+    }
+
+
+publisher = "Springer"
```
Neodia, line 5: This is still being used as an external var

Neodia (approved):

PR #30, opened 2022-05-24
schema for Enhancement

ErnestaP:


Neodia:
I think we should let this PR open until we add the validation. When colliding the schemas with the code we can check more easily if there are problems in the types. This PR should serve as documentation until then. WDYT @ErnestaP @drjova

tests/units/schemas/test_enhancement_schema.py
```diff
@@ -0,0 +1,33 @@
+import json
+
+from schemas.enhancement_schema import EnhancementSchema
+
+
```
Neodia, line 6: What about using pytest parameters in this file ?

tests/units/schemas/test_generic_parser_schema.py
```diff
@@ -0,0 +1,48 @@
+import json
+
+from schemas.generic_parser_schema import GenericParserSchema
+
+
```
Neodia, line 6: Same here

tests/units/schemas/test_parser_schema.py
```diff
@@ -0,0 +1,56 @@
+import json
+
+from schemas.parser_schema import ParserSchema
+
+
```
Neodia, line 6: Same here

ErnestaP:
I think we can merge it since this code is not used at all. Later, we can have a painful time while solving merge conflicts.
Talking about checking is the problems with types- tests are written for that. For now, I'm testing with the input, and output we have from the parser (this data are saved in files). Afterward, we need to change these files, according to what we will get from the parser in QA. Finally, the code will be forced to change as well. We don't need to check it field by field, it is enough to run tests and sees what error it returns, and change code accordingly.

Neodia (approved):

drjova:
@ErnestaP thank you for the code, I believe we are in a stage of investigating if the current setup works well on QA hence I will suggest to keep it open.

pamfilos:
closing since schemas won't be used in the end

PR #32, opened 2022-05-24
Enricher: Added schema and categories

Neodia:


dags/common/enricher.py
```diff
@@ -0,0 +1,95 @@
...
+        def get_arxiv_categories_from_response_xml(xml: ET.Element):
+            xml_namespaces = {
+                "arxiv": "http://arxiv.org/schemas/atom",
+                "w3": "http://www.w3.org/2005/Atom",
+            }
+
+            entries = xml.findall("./w3:entry", namespaces=xml_namespaces)
+            if len(entries) != 1:
+                return []
+            entry = entries[0]
```
drjova: what if `len(entries) == 0`?

dags/common/enricher.py
```diff
@@ -0,0 +1,95 @@
...
+
+            primary_categories = [
+                node.attrib["term"]
+                for node in entry.findall(
+                    "./arxiv:primary_category", namespaces=xml_namespaces
+                )
+            ]
```
drjova: The following doesn't work?
```suggestion
               entry.findall(
                    "./arxiv:primary_category/@term", namespaces=xml_namespaces
                )
```

do we need the namespaces? you can strip them imo

dags/common/enricher.py
```diff
@@ -0,0 +1,95 @@
...
+            if not primary_categories:
+                return []
+            if len(primary_categories) > 1:
+                Logger.error(
+                    "Arxiv returned %d primary categories." % len(primary_categories)
+                )
```
drjova:
```suggestion
                Logger.error(
                    f"Arxiv returned {len(primary_categories)} primary categories." 
```

dags/common/enricher.py
```diff
@@ -0,0 +1,95 @@
...
+
+            return [primary_category] + secondary_categories
+
+        arxiv_id = clean_arxiv(arxiv_id)
+
+        query = f'id:"{arxiv_id}"'
+        response = requests.get(
+            f"http://export.arxiv.org/api/query?search_query={query}"
+        )
+
```
drjova: response.raise_for_status()

dags/common/enricher.py
```diff
@@ -0,0 +1,95 @@
...
+            primary_category = primary_categories[0]
+
+            secondary_categories = [
+                node.attrib["term"]
+                for node in entry.findall("./w3:category", namespaces=xml_namespaces)
+            ]
+            if primary_category in secondary_categories:
+                secondary_categories.remove(primary_category)
+
+            return [primary_category] + secondary_categories
```
drjova: you can create [sets](https://www.geeksforgeeks.org/sets-in-python/) that guarantees no duplicates

dags/common/enricher.py
Neodia: Then return an empty array

drjova: no this will fail, you are checking just for `len(entries) != 1` which means the `len(entries) ==  0` will go to next line and call `entry = entries[0]` which will fail.

```python
In [2]: categories = []

In [3]: len(categories)
Out[3]: 0

In [4]: categories[0]
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
<ipython-input-4-56c311044c1c> in <module>()
----> 1 categories[0]

IndexError: list index out of range
```

dags/common/enricher.py
Neodia: About the attribute : 
```
FAILED tests/units/common/test_enricher.py::test_get_arxiv_categories_arxiv_id - KeyError: '@'
FAILED tests/units/common/test_enricher.py::test_get_arxiv_categories_title - KeyError: '@'
FAILED tests/units/common/test_enricher.py::test_get_arxiv_categories_doi - KeyError: '@'
```

About the namespaces : 
```
FAILED tests/units/common/test_enricher.py::test_get_arxiv_categories_arxiv_id - SyntaxError: prefix 'arxiv' not found in prefix map
FAILED tests/units/common/test_enricher.py::test_get_arxiv_categories_title - SyntaxError: prefix 'arxiv' not found in prefix map
FAILED tests/units/common/test_enricher.py::test_get_arxiv_categories_doi - SyntaxError: prefix 'arxiv' not found in prefix map
```

Neodia: The lib that we're using, `xml.etree`, is a lot more primitive than other libs

dags/common/enricher.py
Neodia: As discussed, the code returns an empty array in case of `len(entries) != 1`

dags/common/enricher.py
Neodia: Done

dags/common/enricher.py
Neodia: Added

dags/common/enricher.py
Neodia: Done

dags/common/enricher.py
```diff
@@ -0,0 +1,111 @@
...
+        arxiv_id = clean_arxiv(arxiv_id)
+
+        search_query = []
+
+        if title:
+            title = (
+                title.replace("-", "?")
+                .encode("ascii", "replace")
+                .decode()
+                .replace("%22", "+")
```
ErnestaP: why here need to replace %22 (which is ") to +?

Neodia: Yes good point, was a leftover from tests

ErnestaP (approved):

dags/common/enricher.py
```diff
@@ -0,0 +1,106 @@
...
+import requests
+from structlog import get_logger
+
+
+class Enricher(object):
+    def __init__(self) -> None:
+        self.logger = get_logger().bind(class_name=type(self).__name__)
+
+    def _get_schema(self):
+        return "http://repo.scoap3.org/schemas/hep.json"
```
drjova: this should not be hardcoded, we will run the code in different environments hence we will have different schema paths i.e. http://repo.qa.scoap3.org

tests/integration/common/test_enricher.py
```diff
@@ -0,0 +1,21 @@
...
+    assert list(set(["hep-th", "hep-ph"])) == enricher._get_arxiv_categories(
+        arxiv_id="2112.01211"
+    )
+
+
+@pytest.mark.vcr
+def test_get_arxiv_categories_title(enricher):
+    assert ["hep-ex"] == enricher._get_arxiv_categories(
+        title="A strategy for a general search for new phenomena using data-derived signal regions and its application within the ATLAS experiment"
+    )
```
drjova, line 24: It's missing an integration test that gives an article and the enricher should give back the article + the `$schema` and `categories`

dags/common/enricher.py
```diff
@@ -0,0 +1,106 @@
...
+        return categories
+
+    def __call__(self, article: Dict):
+        enriched_article = article.copy()
+        enriched_article.update(
+            {
+                "$schema": self._get_schema(),
+                "categories": self._get_arxiv_categories(
+                    article.get("external_id", None)
+                ),
```
drjova: shouldn't update directly the `arxiv_eprints` field?

dags/common/enricher.py
Neodia: I didn't know about that. Done 👍

tests/integration/common/test_enricher.py
Neodia: Done

dags/common/enricher.py
Neodia: Yes good point. Changed 👍

dags/common/enricher.py
```diff
@@ -0,0 +1,113 @@
...
+import requests
+from structlog import get_logger
+
+
+class Enricher(object):
+    def __init__(self) -> None:
+        self.logger = get_logger().bind(class_name=type(self).__name__)
+
+    def _get_schema(self):
+        return os.getenv("REPO_URL", "http://repo.qa.scoap3.org") + "/schemas/hep.json"
```
drjova: this is a challenging one, because we don't have any restriction/checker on how we add url in the env vars. For example we could add either `http://repo.qa.scoap3.org` or `http://repo.qa.scoap3.org/`. It will be safer to do one of the following:

1. Expect the full schema url (which also allow us to give different schemas i.e. inspire or a newer version
2. Build the url using `furl` module

For this specific case I will recommend option 1. WDYT?

dags/common/enricher.py
```diff
@@ -0,0 +1,113 @@
...
+        if arxiv_id is None and title is None and doi is None:
+            raise ValueError(
+                "One of the arxiv_id, title and doi parameters has to be different then None."
+            )
+
+        def clean_arxiv(arxiv):
+            if arxiv is None:
+                return None
+
+            return arxiv.split(":")[-1].split("v")[0].split(" ")[0].strip("\"'")
```
drjova: could you please send me a use case of this code?

dags/common/enricher.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __call__(self, article: Dict):
+        enriched_article = article.copy()
+        enriched_article.update(
+            {
+                "$schema": self._get_schema(),
+                "arxiv_eprints": [
+                    self._set_categories(eprint)
+                    for eprint in enriched_article.get("arxiv_eprints", [])
+                ],
```
drjova, line 110: this structure is wrong according to the `schema`:

```json
{
        "arxiv_eprints": {
            "uniqueItems": true,
            "items": {
                "properties": {
                    "value": {
                        "type": "string",
                        "pattern": "\\d{4}.\\d{4,5}|\\w+-\\w+/\\d+|\\w+/\\d+"
                    },
                    "categories": {
                        "items": {
                            "type": "string"
                        },
                        "type": "array"
                    }
                }
            },
            "type": "array"
        },
}
```

tests/integration/common/test_enricher.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def test_enricher(enricher: Enricher):
+    input_article = {"arxiv_eprints": [{"value": "2112.01211"}]}
+    assert {
+        "arxiv_eprints": [
+            {"value": "2112.01211", "categories": list(set(["hep-th", "hep-ph"]))}
```
drjova: why do we need the set here? Can we just `["hep-th", "hep-ph"]`

drjova (commented):
Thanks, just few more comments!

dags/common/enricher.py
Neodia: `1807.07447v1` this is in one of the test files that we're using.
I did not challenge this code in legacy so let me know if something should be changed

drjova: imo we can simplify this with a simple regex `(\d{4}\.\d{4,5})` this will always get `1807.07447` the numbers between the dot

tests/integration/common/test_enricher.py
Neodia: It's because of the issue with lists order assertion. For other tests I sorted the lists

dags/common/enricher.py
Neodia: I agree with option 1 ! Added 👍

dags/common/enricher.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self) -> None:
+        self.logger = get_logger().bind(class_name=type(self).__name__)
+
+    def _get_schema(self):
+        return os.getenv("REPO_URL", "http://repo.qa.scoap3.org/schemas/hep.json")
+
+    @backoff.on_exception(
+        backoff.expo, requests.exceptions.RequestException, max_time=60, max_tries=3
+    )
+    def _get_arxiv_categories(self, arxiv_id=None, title=None, doi=None):
```
drjova, line 62: @Neodia this function is too large, imo we should split the logic in smaller pieces

dags/common/enricher.py
Neodia: This is a great idea ! Added 👍

dags/common/enricher.py
Neodia: I moved the functions inside it outside, and I remove some unnecessary code. Let me know if it's better

dags/common/enricher.py
```diff
@@ -0,0 +1,109 @@
...
+    def __init__(self) -> None:
+        self.logger = get_logger().bind(class_name=type(self).__name__)
+
+    def _get_schema(self):
+        return os.getenv("REPO_URL", "http://repo.qa.scoap3.org/schemas/hep.json")
+
+    def _clean_arxiv(self, arxiv):
+        if arxiv is None:
+            return None
+        return re.findall(r"(\d{4}\.\d{4,5})", arxiv)[0]
```
drjova: 🙃 this will fail if doesn't find a match, for example:

```

In [0]: re.findall(r"(\d{4}\.\d{4,5})", '2222.000')[0]                                                                                                                                                                                                                                                                                                            
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
```

we could do something like:

```python
try:
    return re.match(r"(\d{4}\.\d{4,5})", '2222.000').group() 
except AttributeError:
    return None
```

dags/common/enricher.py
```diff
@@ -0,0 +1,109 @@
...
+        if response.status_code == 200:
+            xml = ET.fromstring(response.content)
+            categories = self._get_arxiv_categories_from_response_xml(xml)
+            if not categories:
+                self.logger.warning(
+                    f"Could not get arxiv categories for id={arxiv_id} title={title} doi={doi}"
+                )
+        else:
+            self.logger.error(
+                f"Got status_code {response.status_code} from arXiv when looking for categires for id={arxiv_id} title={title} doi={doi}"
```
drjova: are we running black, the line seems more than 88 cols?

tests/integration/common/test_enricher.py
drjova: you can use `assertListEqual` from `unittest` built-in library to avoid all these :)

tests/integration/common/test_enricher.py
```diff
@@ -0,0 +1,31 @@
...
+@pytest.mark.vcr
+def test_get_arxiv_categories_arxiv_id(enricher):
+    ["hep-th", "hep-ph"].sort() == enricher._get_arxiv_categories(
+        arxiv_id="2112.01211"
+    ).sort()
+
+
+@pytest.mark.vcr
+def test_get_arxiv_categories_title(enricher):
+    assert ["hep-ex"].sort() == enricher._get_arxiv_categories(
```
drjova: unnecessary `sort` :)

dags/common/enricher.py
```diff
@@ -0,0 +1,109 @@
...
+        if arxiv_id:
+            params["id_list"] = arxiv_id
+        if title:
+            params[
+                "search_query"
+            ] = f'ti:"{title.replace("-", "?").encode("ascii", "replace").decode()}"'
+
+        request = requests.PreparedRequest()
+        request.prepare_url("http://export.arxiv.org/api/query", params)
+        response = requests.get(request.url)
```
drjova: why not simply 🙃 :

```python
response = requests.get("http://export.arxiv.org/api/query", params=params) 
```

dags/common/enricher.py
```diff
@@ -0,0 +1,109 @@
...
+
+        arxiv_id = self._clean_arxiv(arxiv_id)
+
+        params = {}
+        if arxiv_id:
+            params["id_list"] = arxiv_id
+        if title:
+            params[
+                "search_query"
+            ] = f'ti:"{title.replace("-", "?").encode("ascii", "replace").decode()}"'
```
drjova: we do we need to do all these? 🤔

tests/integration/common/test_enricher.py
Neodia: This would require to change the structure of the tests to add a TestCase class, am I right ?

dags/common/enricher.py
Neodia: I think those were there previously because of my testing. The tests seem to be passing without them so I removed

dags/common/enricher.py
Neodia: Good point 😆

dags/common/enricher.py
Neodia: We are using black yes, it's weird 🤔 
I formatted the lines to be < 88 cols

dags/common/enricher.py
Neodia: Yes good point, updated

tests/integration/common/test_enricher.py
Neodia: Should I change the tests structure to TestCase only to remove the sorts ? Or maybe there is a way without restructuring the tests ?

tests/integration/common/test_enricher.py
```diff
@@ -0,0 +1,41 @@
+import unittest
+
+import pytest
+from common.enricher import Enricher
+
+
+@pytest.fixture
+def test_case():
+    return unittest.TestCase()
+
```
drjova: I would add this in `conftest.py` and do something like:

```python

@pytest.fixture
def test_case_instance():
    return unittest.TestCase()


@pytest.fixture(autouse=True)
def assertListEqual(test_case_instance, first, second):
    return test_case_instance.assertListEqual(first, second)


```
In the tests

```python
@pytest.mark.vcr
 def test_get_arxiv_categories_arxiv_id(enricher):
     assertListEqual(
         ["hep-th", "hep-ph"], enricher._get_arxiv_categories(arxiv_id="2112.01211")
     )

```

dags/__main__.py
```diff
@@ -0,0 +1,45 @@
...
+    json_dump = json.dumps(sorted, ensure_ascii=False)
+
+    try:
+        schema = requests.get("https://repo.scoap3.org/schemas/hep.json").json()
+        jsonschema.validate(sorted, schema)
+    except jsonschema.ValidationError as e:
+        print("PATH", e.message)
+        print("PATH", e.json_path)
+    finally:
+        print(json_dump)
```
drjova: leftover?

dags/common/enricher.py
Neodia: I updated. The replace still seems to be needed or the API doesn't return anything

tests/integration/common/test_enricher.py
Neodia: I pushed a solution that ressembles what you sent. Let me know if that's good 👍

dags/__main__.py
Neodia: It was my local testing file. I added it by mistake. Deleted

drjova (approved):

PR #33, opened 2022-05-30
APS parsing

ErnestaP:
some fields are different from Springer output. 
* APS JSON doesn't have collaboration, should I add it from the affiliations organization?
*  In the old code, the authors' structure was different. Now I adapted to Springer one).
* APS has one more field- **field_categories**. What should I do with it? keep it, parse it into a different format?

### DIFF
--- a/.github/workflows/test-and-build.yml
+++ b/.github/workflows/test-and-build.yml
@@ -23,9 +23,10 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
+          pip install --upgrade pip
+          pip install --no-cache-dir --upgrade setuptools
+          pip install --no-cache-dir --upgrade wheel
           pip install -r requirements-airflow.txt -r requirements.txt -r requirements-test.txt
-
       - name: Create S3 Buckets
         run: make buckets
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,6 +6,9 @@ ENV AIRFLOW_UID=501
 COPY requirements.txt ./requirements.txt
 COPY requirements-test.txt ./requirements-test.txt
 COPY dags ./dags
-
+RUN pip install --upgrade pip &&\
+    pip install --no-cache-dir --upgrade setuptools &&\
+    pip install --no-cache-dir --upgrade wheel &&\
+    pip install --no-cache-dir --user -r requirements.txt -r requirements-test.txt
 USER airflow
-RUN pip install --no-cache-dir --user -r requirements.txt -r requirements-test.txt
+
--- a/dags/aps/parser.py
+++ b/dags/aps/parser.py
@@ -1,0 +1,135 @@
+import re
+
+from common.parsing.json_extractors import CustomExtractor, NestedValueExtractor
+from common.parsing.parser import IParser
+from inspire_utils.record import get_value
+from structlog import get_logger
+
+
+class APSParser(IParser):
+    def __init__(self) -> None:
+        self.logger = get_logger().bind(class_name=type(self).__name__)
+        article_type_mapping = {
+            "article": "article",
+            "erratum": "erratum",
+            "editorial": "editorial",
+            "retraction": "retraction",
+            "essay": "other",
+            "comment": "other",
+            "letter-to-editor": "other",
+            "rapid": "other",
+            "brief": "other",
+            "reply": "other",
+            "announcement": "other",
+            "nobel": "other",
+        }
+
+        extractors = [
+            NestedValueExtractor("dois", json_path="identifiers.doi"),
+            NestedValueExtractor(
+                "journal_doctype",
+                json_path="articleType",
+                extra_function=lambda x: article_type_mapping[x],
+            ),
+            NestedValueExtractor("page_nr", json_path="numPages"),
+            NestedValueExtractor(
+                "arxiv_eprints",
+                json_path="identifiers.arxiv",
+                extra_function=lambda x: {
+                    "value": re.sub("arxiv:", "", x, flags=re.IGNORECASE)
+                },
+            ),
+            NestedValueExtractor("abstract", json_path="abstract.value"),
+            NestedValueExtractor("title", json_path="title.value"),
+            CustomExtractor("authors", self._form_authors),
+            NestedValueExtractor("journal_title", json_path="journal.name"),
+            NestedValueExtractor("journal_issue", json_path="issue.number"),
+            NestedValueExtractor("journal_volume", json_path="volume.number"),
+            NestedValueExtractor(
+                "journal_year", json_path="date", extra_function=lambda x: int(x[:4])
+            ),
+            NestedValueExtractor("date_published", json_path="date"),
+            NestedValueExtractor(
+                "copyright_holder",
+                json_path="rights.copyrightHolders",
+                extra_function=lambda x: x[0]["name"],
+            ),
+            NestedValueExtractor("copyright_year", json_path="rights.copyrightYear"),
+            NestedValueExtractor(
+                "copyright_statement", json_path="rights.rightsStatement"
+            ),
+            NestedValueExtractor(
+                "licenses",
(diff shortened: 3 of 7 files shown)


dags/aps/utils.py
```diff
@@ -52,3 +52,12 @@ def trigger_file_processing(ids_and_articles):
             replace_microseconds=False,
         )
     return ids_and_articles
+
+
+def get_nested(root, *keys):
```
Neodia: This seems to be used in the common library so it should not be here

dags/common/parsing/json_extractors.py
```diff
@@ -0,0 +1,30 @@
+from aps.utils import get_nested
+from common.parsing.extractor import IExtractor
+
+
+class ValueExtractor(IExtractor):
```
Neodia: I find this class' name a bit confusing. Why do we have multiple keys ? Are we setting a value from multiple keys ? I understood that we are getting data from nested elements, but I had to dive deep in the code to understand. IMO we should change this class to extract only the value (IE one destination and one source) and add a class named "NestedValueExtractor". Instead of `keys` I would name this `json_path`. What do you think ?

dags/aps/parser.py
```diff
@@ -0,0 +1,108 @@
...
+            "reply": "other",
+            "announcement": "other",
+            "nobel": "other",
+        }
+
+        extractors = [
+            ValueExtractor("dois", keys=("identifiers", "doi")),
+            ValueExtractor(
+                "journal_doctype",
+                keys=("articleType",),
```
Neodia: @drjova Ernesta and I had a conversation about the type here. I'm used to use lists instead but I saw on the internet that if the data doesn't change, tuples are more memory efficient. What's your take on this ?

drjova: 😅 indeed, tuples are more memory efficient, but in our case, it doesn't make any difference because we don't have big arrays. Hence, it doesn't affect the result.

Also, just for reference, the main difference between a list and a tuple is that list is mutable, and a tuple is immutable. More details [here](https://www.geeksforgeeks.org/python-difference-between-list-and-tuple/)

dags/aps/utils.py
drjova: we have a tested library in inspire [inspire-utils](https://pypi.org/project/inspire-utils/) which we can install and use the `from inspire_utils.record import get_value` function which supports `get_value(data, 'classificationSchemes.subjectAreas')` notation

dags/aps/parser.py
```diff
@@ -0,0 +1,136 @@
...
+                    {
+                        "full_name": author["name"],
+                        "given_names": author["firstname"],
+                        "surname": author["surname"],
+                        "affiliations": self._get_affiliations(
+                            article, author["affiliationIds"]
+                        ),
+                    }
+                )
+        return {"authors": parsed_authors}
```
Neodia:
```suggestion
        authors = article["authors"]
        parsed_authors = []
        for author in authors:
            if author["type"] == "Person":
                parsed_authors.append(
                    {
                        "full_name": author["name"],
                        "given_names": author["firstname"],
                        "surname": author["surname"],
                        "affiliations": self._get_affiliations(
                            article, author["affiliationIds"]
                        ),
                    }
                )
        return {"authors": parsed_authors}
```
```suggestion
        authors = article["authors"]
        parsed_authors = [
             {
                        "full_name": author["name"],
                        "given_names": author["firstname"],
                        "surname": author["surname"],
                        "affiliations": self._get_affiliations(
                            article, author["affiliationIds"]
                        ),
               }
               for author in article['authors']
               if author["type"] == "Person"
        ]
        return {"authors": parsed_authors}
```

dags/aps/parser.py
```diff
@@ -0,0 +1,136 @@
...
+        parsed_affiliations = []
+        for affiliation in affiliations:
+            if affiliation["id"] in affiliationIds:
+                value = affiliation["name"]
+                organization = (",").join(value.split(",")[:-1])
+                country = value.split(", ")[-1:]
+                parsed_affiliations.append(
+                    {"value": value, "organization": organization, "country": country}
+                )
+        return parsed_affiliations
```
Neodia:
```suggestion
        affiliations = article["affiliations"]
        parsed_affiliations = [
            {"value": affiliation["name"], "organization": (",").join(value.split(",")[:-1]), "country": value.split(", ")[-1:]}
            for affiliation in article["affiliations"]
            if affiliation["id"] in affiliationIds
        ]
        return parsed_affiliations
```

Neodia:
```suggestion
        affiliations = article["affiliations"]
        parsed_affiliations = [
            {"value": affiliation["name"], "organization": (",").join(value.split(",")[:-1]), "country": value.split(", ")[-1:]}
            for affiliation in article["affiliations"]
            if affiliation["id"] in affiliationIds
        ]
        return parsed_affiliations
```

Dockerfile
```diff
@@ -6,6 +6,11 @@ ENV AIRFLOW_UID=501
 COPY requirements.txt ./requirements.txt
 COPY requirements-test.txt ./requirements-test.txt
 COPY dags ./dags
-
+RUN pip install wheel
+RUN python setup.py bdist_wheel 
+RUN pip install --no-cache-dir --upgrade setuptools && \
+    pip install --no-cache-dir --upgrade wheel && \
+    pip install --no-cache-dir --upgrade pip==22.0.4 && \
+    pip install --no-cache-dir --user -r requirements.txt -r requirements-test.txt
```
drjova: you already have pip install wheel :) and why hardcode the pip version?

.github/workflows/test-and-build.yml
```diff
@@ -23,9 +23,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          pip install -r requirements-airflow.txt -r requirements.txt -r requirements-test.txt
-
+         pip install --no-cache-dir --upgrade setuptools
+         pip install -r requirements-airflow.txt -r requirements.txt -r requirements-test.txt
```
drjova: 🙃 why you have changed it? do we have to update the setuptools? if so, please add `pip install --upgrade pip` as well before

Dockerfile
```diff
@@ -6,6 +6,9 @@ ENV AIRFLOW_UID=501
...
 COPY requirements-test.txt ./requirements-test.txt
 COPY dags ./dags
-
+RUN pip install wheel
+RUN python setup.py bdist_wheel 
+RUN pip install --no-cache-dir --upgrade setuptools && \
+    pip install --no-cache-dir --user -r requirements.txt -r requirements-test.txt
 USER airflow
-RUN pip install --no-cache-dir --user -r requirements.txt -r requirements-test.txt
+
```
drjova: Adding multiple `RUN` adds layer to docker hence it makes the image build slower. You can combine all the commands into one. What I don't understand is why we have to install wheel, run  python setup.py bdist_wheel  and upgrade setuptools? This question also applies to the workflows file.

dags/aps/parser.py
ErnestaP: I couldn't commit this change, because it has some mistakes. However, I did a change regarding to this manually :) thank you

Dockerfile
ErnestaP: I don't, it was one of the solutions I found. Looks like I overdid 😅

.github/workflows/test-and-build.yml
ErnestaP: yes, we need to update setuptools and wheel, because oterwise inspirehep utils cannot find whee, you can see [here](https://github.com/cern-sis/workflows/runs/6686866925?check_suite_focus=true):
```
      WARNING: The wheel package is not available.
```

Dockerfile
ErnestaP: it was an overdone thing, in order to solve a problem. We don't need to install them, we need just upgrade

Neodia (approved):

PR #34, opened 2022-06-01
Enricher: Final Steps

Neodia:


dags/common/cleanup.py
```diff
@@ -51,3 +51,22 @@ def clean_all_affiliations_for_author(input):
...
+    [obj.pop(field, None) for field in fieldnames]
+    return obj
+
+
+def remove_orcid_prefix(obj):
+    orcid_prefixes = ("orcid:", "https://orcid.org/", "http://orcid.org/")
+    for author in obj.get("authors", ()):
+        for orcid_prefix in orcid_prefixes:
+            if "orcid" in author and author["orcid"].lower().startswith(orcid_prefix):
+                author["orcid"] = author["orcid"][len(orcid_prefix) :]
```
drjova: we can optimise it a bit, to filter out authors without orcid
```suggestion
    for author in obj.get("authors", ()):
        if "orcid" not in author:
            continue
        for orcid_prefix in orcid_prefixes:
            if author["orcid"].lower().startswith(orcid_prefix):
                author["orcid"] = author["orcid"][len(orcid_prefix) :]
```

drjova (approved):
Just the comment

dags/common/cleanup.py
Neodia: Done

PR #37, opened 2022-06-21
Full APS file processing

ErnestaP:
* Merged APS files parsing, generic parsing, enhancement, and enrichment in one dag.
* Added tests
* Fixed some inconsistencies
* ref: https://github.com/cern-sis/issues-scoap3/issues/71


dags/aps/aps.py
```diff
@@ -27,18 +27,21 @@ def save_json_in_s3(dates: dict, repo: IRepository = APSRepository()):
...
+        return None
 
     @task()
     def trigger_file_processing(key, repo: IRepository = APSRepository()):
         if key is None:
             logging.warning("No new files were downloaded to s3")
             return
         ids_and_articles = split_json(repo, key)
-        trigger_file_processing_(ids_and_articles)
+        return trigger_file_processing_(ids_and_articles)
```
drjova: do you need the return here? and there is something wrong with the name, it has `_` at the end

dags/aps/aps_process_file.py
```diff
@@ -1,18 +1,59 @@
...
+    validate(enriched_file_copy, schema)
 
 
 @dag(start_date=airflow.utils.dates.days_ago(0))
 def aps_process_file():
     @task()
-    def start():
-        return None
+    def parse(**kwargs):
+        if kwargs["params"] and "article" in kwargs["params"]:
```
drjova:
```suggestion
        if 'params' in kwargs and "article" in kwargs["params"]:
```

dags/aps/aps_process_file.py
```diff
@@ -1,18 +1,59 @@
...
+            print(kwargs["params"]["article"])
```
drjova: do we need the print?

dags/aps/aps_process_file.py
```diff
@@ -1,18 +1,59 @@
...
+            article = json.loads(kwargs["params"]["article"])
+            return parse_aps(article)
+
+    @task()
+    def enchance(parsed_file):
+        return parsed_file and enhance_aps(parsed_file)
```
drjova, line 41: do we return bool?

dags/aps/aps_process_file.py
```diff
@@ -1,18 +1,59 @@
...
+
+    @task()
+    def enrich(enhanced_file):
+        return enhanced_file and enrich_aps(enhanced_file)
```
drjova, line 45: do we return a bool?

dags/aps/parser.py
```diff
@@ -46,53 +50,53 @@ def __init__(self) -> None:
...
+                "journal_year",
+                json_path="date",
+                extra_function=lambda x: int(x[:4] if (len(x) >= 4) else 0000),
             ),
             NestedValueExtractor("date_published", json_path="date"),
             NestedValueExtractor(
                 "copyright_holder",
                 json_path="rights.copyrightHolders",
-                extra_function=lambda x: x[0]["name"],
+                extra_function=lambda x: x[0]["name"] if len(x) >= 1 else "",
```
drjova, line 61:
```suggestion
                extra_function=lambda x: x[0]["name"] if len(x) >= 0 else "",
```
?

dags/aps/utils.py
```diff
@@ -48,7 +57,8 @@ def trigger_file_processing(ids_and_articles):
         trigger_dag.trigger_dag(
             dag_id="aps_process_file",
             run_id=data["id"],
-            conf=json.dumps(data["article"]),
+            conf={"article": json.dumps(data["article"])},
             replace_microseconds=False,
         )
-    return ids_and_articles
+    print(type(json.dumps(data["article"])))
```
drjova: we don't need the print

tests/units/aps/test_aps_parser.py
```diff
@@ -165,59 +158,60 @@ def parsed_articles(parser, articles):
...
-                            "name": "{0}.xml".format("10.1103/PhysRevB.103.165408"),
-                            "filetype": "xml",
-                        },
-                    ]
-                },
-            ],
-            "extra_data",
-            id="test_extra_data",
-        ),
+        # THIS IS A TEST FOR EXTRA DATA
```
drjova, line 168: we shouldn't leave comments :)

dags/aps/aps.py
ErnestaP: I just called two different functions with the same name, so I just added `_`. Will rename it. 
trigger_file_processing_ returns because in the xcom section in UI we can see which articles were passed to the following DAG. Useful for following that is happening

dags/aps/aps_process_file.py
ErnestaP: No

dags/aps/aps_process_file.py
ErnestaP: No, we return dict or None

dags/aps/aps_process_file.py
ErnestaP: No, we return dict or None

dags/aps/parser.py
ErnestaP: I'm afraid the suggestion is wrong. For example, we have:
```
x=[{'name':  'Copy Holder Name'}]
```
the length of this array is 1, so we can take the first item (x[0]) name.
If
```
x=[]
```
the length of this array is 0, and when we will try to take the value of x[0]['name'] it will give us an error of:
```
IndexError: list index out of range
```

tests/units/aps/test_aps_parser.py
ErnestaP: There is one more comment. It was just a question and reminder about extra data 😅:
https://github.com/cern-sis/workflows/blob/full_aps_proccessing/dags/aps/parser.py#L73-L74

dags/aps/aps_process_file.py
```diff
@@ -1,18 +1,58 @@
...
+def enhance_aps(parsed_file):
+    return Enhancer()("APS", parsed_file)
+
+
+def enrich_aps(enhanced_file):
+    return Enricher()(enhanced_file)
+
+
+def aps_validate_record(enriched_file):
+    enriched_file_copy = enriched_file.copy()
```
Neodia: Why are you creating a copy ?

tests/units/aps/test_aps_parser.py
ErnestaP: Deleted comments

dags/aps/aps_process_file.py
ErnestaP: No need, you're right, changed it :)

dags/aps/aps.py
```diff
@@ -27,22 +31,25 @@ def save_json_in_s3(dates: dict, repo: IRepository = APSRepository()):
         rest_api = APSApiClient(
             base_url=os.getenv("APS_API_BASE_URL", "http://harvest.aps.org")
         )
-        articles_metadata = str.encode(
-            json.dumps(rest_api.get_articles_metadata(parameters))
-        )
-        return save_file_in_s3(data=articles_metadata, repo=repo)
+        articles_metadata = rest_api.get_articles_metadata(parameters)
+        if articles_metadata is not None:
+            articles_metadata = str.encode(
```
drjova: I still don't understand why encoding :/

ErnestaP: Otherwise, I'm getting an error, when I'm trying to write it to the file:
` TypeError: a bytes-like object is required, not 'str'`
Code, where the error is raised: [here](https://github.com/cern-sis/workflows/blob/full_aps_proccessing/dags/aps/utils.py#L34)

drjova: `get_articles_metadata` already returning [json](https://github.com/cern-sis/workflows/blob/e7d33f397a72df5abd87d8cd5bdb4ed7c7a30111/dags/aps/aps_api_client.py#L26) why you don't save it directly?

tests/units/aps/test_aps_file_processing.py
```diff
@@ -0,0 +1,43 @@
...
+@fixture
+def enrich_article(enhance_articles):
+    return [enrich_aps(article) for article in enhance_articles]
+
+
+def test_aps_validated_record(enrich_article):
+    try:
+        [aps_validate_record(article) for article in enrich_article]
+    except (Exception):
+        fail()
```
drjova: what this is testing?

dags/aps/utils.py
```diff
@@ -3,20 +3,26 @@
...
-from airflow.models import Variable
 
 
-def set_APS_harvesting_interval(repo):
-    today = date.today()
-    aps_fetching_from_date = repo.find_the_last_uploaded_file_date()
-    aps_fetching_until_date = today.strftime("%Y-%m-%d")
-    Variable.set(
-        "aps_fetching_from_date", aps_fetching_from_date or aps_fetching_until_date
+def set_APS_harvesting_interval(repo, **kwargs):
```
Neodia: There are other publishers that  pull from API right ? If that's the case I think this function could go to common folder. What do you think ?

tests/units/aps/test_aps_file_processing.py
ErnestaP: is the article is correct, after all steps of workflow: it runs validation on article

drjova: hmm why it needs the try/except?

dags/aps/aps.py
ErnestaP: solved ☺️

tests/units/aps/test_aps_file_processing.py
ErnestaP: thinking more, I see your point, we don't need it

dags/aps/utils.py
ErnestaP: Good idea, moved it

Neodia (approved):

drjova (approved):

dags/aps/parser.py
```diff
@@ -25,19 +25,23 @@ def __init__(self) -> None:
...
         extractors = [
-            NestedValueExtractor("dois", json_path="identifiers.doi"),
+            NestedValueExtractor(
+                "dois", json_path="identifiers.doi", extra_function=lambda x: [x]
+            ),
             NestedValueExtractor(
                 "journal_doctype",
                 json_path="articleType",
-                extra_function=lambda x: article_type_mapping[x],
+                extra_function=lambda x: article_type_mapping[x] if x in article_type_mapping else 'other' ,
```
drjova:
```suggestion
                extra_function=lambda x: article_type_mapping.get(x, 'other'),
```

dags/common/parsing/json_extractors.py
```diff
@@ -9,15 +9,17 @@ def __init__(
         json_path,
         required=True,
         extra_function=lambda s: s,
+        default="",
```
drjova: it's not a good practise to have mutable args in the signature https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

drjova (approved):
LGTM 🚀

PR #39, opened 2022-06-29
APS api parameters fix, according old configuration

ErnestaP:
* ref: https://github.com/cern-sis/kubernetes/blob/scoap3-prod/scoap3/base/scoap3.cfg#L178-L187


dags/aps/aps_params.py
```diff
@@ -25,6 +23,5 @@ def get_params(self) -> dict:
             "date": self.date,
             "journals": self.journals,
             "sets": self.sets,
-            "per_page": self.per_page,
```
drjova, line 28: what's the default per_page?

ErnestaP: max 100, I will put 100 back as default
doc: http://harvest.aps.org/docs/harvest-api#endpoints

tests/integration/aps/cassettes/test_get_articles_metadata.yaml
```diff
@@ -1,85 +1,84 @@
...
+      Accept:
+      - '*/*'
+      Accept-Encoding:
+      - gzip, deflate
+      Connection:
+      - keep-alive
+      User-Agent:
+      - python-requests/2.27.1
+    method: GET
+    uri: http://harvest.aps.org/v2/journals/articles?from=2021-04-10&until=2022-04-10&date=modified&sets=scoap3&per_page=1
```
drjova, line 14: the request here has `per_page=1`

ErnestaP: yes, because the test has it. I passed not default value here"
https://github.com/cern-sis/workflows/blob/main/tests/integration/aps/test_aps_api_client.py#L17

drjova (approved):

PR #42, opened 2022-07-05
APS parsing improvement

ErnestaP:
* hardcoded collection values;
* chnaged sets parameter to set
* removed parameters from get_xml_file and get_pdf_file, because they are not needed
* ref: https://github.com/cern-sis/issues-scoap3/issues/75


dags/aps/parser.py
```diff
@@ -68,7 +68,7 @@ def __init__(self) -> None:
                 "license",
                 json_path="rights.licenses",
             ),
-            NestedValueExtractor("collections", json_path="HEP.Citeable.Published"),
+            CustomExtractor("collections", extraction_function=lambda x: ["HEP", "Citeable", "Published"]),
```
Neodia: You have to see lambdas as nothing more than a function : If no param is needed you don't put it : 
```python3
lambda: ["HEP", "Citeable", "Published"]
```

ErnestaP: For now it won't work like this.
I need to change the code here: https://github.com/cern-sis/workflows/blob/main/dags/common/parsing/json_extractors.py#L29-L30

it expects a variable, which at this moment is x in lambda

ErnestaP: interesting, try to fix it and getting an error:
`TypeError: <lambda>() takes 0 positional arguments but 1 was given`

Neodia: Ohh yes indeed, because we generalized its usage there always is a parameter, even if it's not used

Neodia (approved):

dags/aps/parser.py
```diff
@@ -64,11 +65,11 @@ def __init__(self) -> None:
             NestedValueExtractor(
                 "copyright_statement", json_path="rights.rightsStatement"
             ),
-            NestedValueExtractor(
+            CustomExtractor(
                 "license",
-                json_path="rights.licenses",
+                extraction_function=lambda x: self._get_licenses(x['rights']['licenses'][0])
```
drjova: does it always have licenses? `[0]` might fail. Also we have a license function on Springer, please check if we can reuse it.

dags/common/parsing/json_extractors.py
```diff
@@ -26,5 +26,5 @@ def __init__(self, destination, extraction_function) -> None:
         self.destination = destination
         self.extraction_function = extraction_function
 
-    def extract(self, article: dict):
+    def extract(self, article: dict = {}):
```
drjova: we shouldn't use mutable for args

tests/units/aps/test_aps_parser.py
```diff
@@ -158,14 +158,14 @@ def parsed_articles(parser, articles):
...
-                [{"url": "http://link.aps.org/licenses/aps-default-license"}],
+                [{'license': 'CC-LICENSES-aps-default-license', "url": "http://link.aps.org/licenses/aps-default-license"}],
+                [{'license': 'CC-LICENSES-aps-default-license', "url": "http://link.aps.org/licenses/aps-default-license"}],
             ],
             "license",
             id="test_license",
         ),
         pytest.param([[], []], "field_categories", id="test_field_categories"),
-        # THIS IS A TEST FOR EXTRA DATA
+        pytest.param([['HEP', 'Citeable', 'Published'], ['HEP', 'Citeable', 'Published']], "collections", id="test_collections"),
```
drjova: black/pre-commit?

dags/aps/parser.py
ErnestaP: Springer parses XML, so it's a string. For APS it's JSON, and it's always in the array.
However, I can add catch/except block

drjova: It doesn't matter. We can modify the license function to accept parameters that can be used for both. Anyway, for both publishers, we should return a JSON

drjova: The logic is pretty similar https://github.com/cern-sis/workflows/blob/main/dags/springer/parser.py#L240-L266

ErnestaP: I see your point, but I need to extract it from JSON, not from XML, in the end, I cannot reuse it.
The license mappings I took from scoap3/extractors: https://github.com/SCOAP3/hepcrawl/blob/c10a838e94b6457356aabb9bb370eef35ec2dc60/hepcrawl/mappings.py#L189-L193

There could be more licenses, not just CC-CC, CC-BY but also arXiv-.
Also, for Springer the URL has to be constructed, while for APS I'm getting it from JSON.
The license. license is extracted from the value in XML, while in APS I have to do it from URL.

In the end, there are two different values: Springer is getting license.license from XML, and we have to construct URL. While with APS is the opposite: URL is given and  license.license  has to be constructed

drjova: We can move the extraction parts on the publisher's parser and have a common function to return the licenses

dags/common/utils.py
```diff
@@ -11,10 +14,31 @@ def set_harvesting_interval(repo, **kwargs):
...
+                )
+                return {"url": url, "license": f"CC-{license_type}-{version}"}
+    else:
+        logger.warning(
+            "Licence is not given, or missing arguments. Returning default license"
+        )
+        return {
+            "license": "CC-BY-3.0",
+            "url": "https://creativecommons.org/licenses/by/3.0",
+        }
```
```suggestion
def construct_license(url, license_type, version):
    if url and license_type and version:
        for license in LICENSES.keys():
            if license in url:
                return {"url": url, "license": f"{LICENSES[license]}{version}"}
 
          logger.warning(
              f"Given licence with parameters license_type={license_type} version={version} is not in the mappings"
          )
          return {"url": url, "license": f"CC-{license_type}-{version}"}

  logger.warning(
      "Licence is not given, or missing arguments. Returning default license"
  )
  return {
      "license": "CC-BY-3.0",
      "url": "https://creativecommons.org/licenses/by/3.0",
  }
```

drjova: @ErnestaP did we ask Anne for licenses?

dags/aps/parser.py
```diff
@@ -102,8 +108,8 @@ def _get_affiliations(self, article, affiliationIds):
         ]
         return parsed_affiliations
 
-    def _get_field_categories(self, article):
-        return [
+    def _get_field_categories(self, article):  # chech the order!!!
```
drjova:
```suggestion
    def _get_field_categories(self, article):
```

dags/common/utils.py
```diff
@@ -11,10 +12,17 @@ def set_harvesting_interval(repo, **kwargs):
             "start_date": kwargs["params"]["start_date"],
             "until_date": kwargs["params"]["until_date"],
         }
-    start_date = repo.find_the_last_uploaded_file_date()
-    until_date = date.today()
-
+    start_date = datetime.datetime.strptime(
+        repo.find_the_last_uploaded_file_date(), "%Y-%m-%d"
```
drjova: we are already formatting the date in `find_the_last_uploaded_file_date`

dags/common/utils.py
```diff
@@ -11,10 +12,17 @@ def set_harvesting_interval(repo, **kwargs):
...
+    ).date()
+    until_date = datetime.date.today()
```
```suggestion
    until_date = datetime.date.today().strftime("%Y-%m-%d")
```

dags/common/utils.py
ErnestaP:
```
start_date = datetime.datetime.strptime(
        repo.find_the_last_uploaded_file_date(), "%Y-%m-%d"
    ).date()
```  
is a type date
while
```
"start_date": (start_date or until_date).strftime("%Y-%m-%d"),
        "until_date": until_date.strftime("%Y-%m-%d")
``` 
I'm stringify it

drjova: we don't need all these date transformations, let's stick to one. Preferably always use `strftime` since we always need str date for our functions.

dags/common/utils.py
```diff
@@ -12,9 +13,14 @@ def set_harvesting_interval(repo, **kwargs):
             "until_date": kwargs["params"]["until_date"],
         }
     start_date = repo.find_the_last_uploaded_file_date()
-    until_date = date.today()
-
+    until_date = datetime.date.today().strftime("%Y-%m-%d")
```
drjova, line 16: there is  a potential bug for `start_date` https://github.com/cern-sis/workflows/blob/e7d33f397a72df5abd87d8cd5bdb4ed7c7a30111/dags/aps/repository.py#L29 the first time we run it, if there are no s3 object it will return `None`

ErnestaP: no. it will put today's date

drjova (approved):

PR #48, opened 2022-08-10
Hindawi

ErnestaP:


dags/common/utils.py
```diff
@@ -12,7 +12,12 @@ def set_harvesting_interval(repo, **kwargs):
             "start_date": kwargs["params"]["start_date"],
             "until_date": kwargs["params"]["until_date"],
         }
-    start_date = repo.find_the_last_uploaded_file_date()
+    start_date = (
+        "params" in kwargs
+        and "start_date" in kwargs["params"]
+        and kwargs["params"]["start_date"]
```
MJedr: It is a bit odd to me cause it will return boolean and `find_the_last_uploaded_file_date` will return a date.
Did you mean something like 
`kwargs["params"]["start_date"] if kwargs.get('params', {}).get('start_date')`?

dags/hindawi/hindawi.py
```diff
@@ -0,0 +1,46 @@
...
+    def save_xml_in_s3(dates: dict, repo: IRepository = HindawiRepository()):
+        parameters = HindawiParams(
+            from_date=dates["start_date"],
+            until_date=dates["until_date"],
+        ).get_params()
+        rest_api = HindawiApiClient(
+            base_url=os.getenv("HINDAWI_API_BASE_URL", "https://www.hindawi.com")
+        )
+        articles_metadata = rest_api.get_articles_metadata(parameters)
+        if articles_metadata is not None:
```
MJedr: `if articles_metadata:` should be enough I think

dags/hindawi/hindawi.py
```diff
@@ -0,0 +1,46 @@
...
+            return save_file_in_s3(data=articles_metadata, repo=repo)
+
+    @task()
+    def trigger_files_processing(key, repo: IRepository = HindawiRepository()):
+        if key is None:
```
MJedr: `if not key` might work and would be better

dags/hindawi/hindawi.py
```diff
@@ -0,0 +1,46 @@
...
+            return
+        ids_and_articles = split_xmls(repo, key)
+        return trigger_file_processing_DAG(ids_and_articles)
+
+    intervals = set_fetching_intervals()
+    key = save_xml_in_s3(intervals)
+    trigger_files_processing(key)
+
+
+Hindawi_download_files_dag = hindawi_fetch_api()
```
MJedr: `hindawi_download_files_dag` imo should be lowercased

dags/hindawi/hindawi_api_client.py
```diff
@@ -0,0 +1,40 @@
...
+        self.base_url = base_url or getenv(
+            "HINDAWI_API_BASE_URL", "https://www.hindawi.com"
+        )
+        self.path_segments = path_segments or ["oai-pmh", "oai.aspx"]
+
+    def get_articles_metadata(self, parameters, doi=None):
+        file_is_empty = True
+        all_path_segments = self.path_segments
+        if doi is not None:
+            all_path_segments = self.path_segments + [doi]
```
MJedr: maybe just `self.path_segments.append(doi)`

dags/hindawi/hindawi_api_client.py
```diff
@@ -0,0 +1,40 @@
...
+        request = Request(
+            base_url=self.base_url,
+            path_segments=all_path_segments,
+            parameters=parameters,
+        )
+        response_content = request.get_response_xml()
+        for tag in ElementTree.fromstring(response_content):
+            if "ListRecords" in tag.tag:
+                file_is_empty = False
```
MJedr: you can use `break` when you find `ListRecords` in a tag

dags/hindawi/hindawi_file_processing.py
```diff
@@ -0,0 +1,16 @@
...
+
+import airflow
+from airflow.decorators import dag, task
+
+
+@dag(start_date=airflow.utils.dates.days_ago(0))
+def hindawi_file_processing():
+    @task()
+    def parse(**kwargs):
+        if "params" in kwargs and "article" in kwargs["params"]:
```
MJedr: `if params in kwargs.get('params')`

dags/hindawi/hindawi_file_processing.py
```diff
@@ -0,0 +1,16 @@
...
+            article = json.loads(kwargs["params"]["article"])
+            return "Nope"
```
MJedr: why you return this? 😄

dags/hindawi/utils.py
```diff
@@ -0,0 +1,46 @@
+import uuid
+from datetime import date, datetime
+from io import BytesIO
+from xml.etree import ElementTree
+
+from airflow.api.common import trigger_dag
+
+
+def save_file_in_s3(data, repo):
+    if bool(data):
```
MJedr:
```
if not data:
    return
<then rest of the code>
```

dags/hindawi/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        date_today = date.today().strftime("%Y-%m-%d")
+        prefix = datetime.now().strftime("%Y-%m-%dT%H:%M")
```
MJedr: since you have a date in `date_today`, do you need a date also in prefix?

dags/hindawi/utils.py
```diff
@@ -0,0 +1,46 @@
...
+        repo.save(key, byte_file)
+        return key
+
+
+def split_xmls(repo, key):
+    ids_and_records = []
+    file = repo.find_by_id(key)
+    xml_string = file.getvalue().decode("utf-8")
+    for tag in ElementTree.fromstring(xml_string):
+        if "ListRecords" in tag.tag:
```
MJedr:
```
if "ListRecords" in tag.tag:
    continue
```

dags/hindawi/utils.py
```diff
@@ -0,0 +1,46 @@
...
+            for record in tag:
+                if "record" in record.tag:
```
MJedr:
```
if "record" not in record.tag:
    continue
```

dags/hindawi/utils.py
```diff
@@ -0,0 +1,46 @@
...
+                        {
+                            "id": id,
+                            "record": ElementTree.tostring(record).decode("utf-8"),
+                        }
+                    )
+    return ids_and_records
+
+
+def trigger_file_processing_DAG(ids_and_records):
+    if len(ids_and_records) > 0:
```
MJedr:
```
if len(ids_and_records) > 0:
    continue
```

dags/common/utils.py
ErnestaP: no, is not a boolean which is returned here.
kwargs["params"]["start_date"] is string. 
Here what I am doing is: 
1. Checking is params even in kwargs for this task. It can be that kwargs will be empty if we won't pass anything before triggering the task. How to pass it? In UI of airflow, before triggering the run (and run consists of tasks)
2. Checking if `start_date` is in kwargs['params']. We can pass whichever value there, not just `start_date`
3. If we have `params`, and `params['start_date']` defined, I return the value of `params['start_date']`. For example: 

```python 
params['start_date'] = "2020-08-08"
'start_date' in params and params['start_date'] == "2020-08-08"
```

dags/hindawi/hindawi_file_processing.py
ErnestaP: this is just for ref :D will be changed in the next PR. This one focuses just on downloading files ( [dags/hindawi/hindawi.py](https://github.com/cern-sis/workflows/pull/48/files/c6bbf066dbf9297fc2171f9da45ef1102c1ea67c#diff-4ec5509cda5b0af6ea04283614f6c339b9fa18033eb5669f9e77f328217886d3))

dags/hindawi/utils.py
ErnestaP: yes, because the same day we can have more than one file downloaded. Let's say we downloaded 5 files today.
So, the "folder" in S3 will be the date "2022-08-10", however, the key of the file itself shows also the time when it was downloaded. In the end,  structure looks like this:
2022-08-10/2022-08-10T17:33.xml
2022-08-10/2022-08-10T18:36.xml
2022-08-10/2022-08-10T19:00.xml
2022-08-10/2022-08-10T21:01.xml
2022-08-10/2022-08-10T23:33.xml

dags/hindawi/utils.py
ErnestaP: This one I didn't really get it 😅 could you explain it a little bit more?

dags/hindawi/utils.py
ErnestaP: same

dags/hindawi/utils.py
ErnestaP: annnddd same

dags/common/utils.py
MJedr: So I would suggest doing `kwargs.get('params', {}).get('start_date')` instead
```
        "params" in kwargs
        and "start_date" in kwargs["params"]
        and kwargs["params"]["start_date"]
```
to me, it's a bit more readable with just one statement

dags/hindawi/utils.py
MJedr: sorry, I meant:
```
if "ListRecords" not in tag.tag:
    continue
```
to remove one indentation level. Then you can continue the rest of the code in the loop without an indentation, so
```
    for tag in ElementTree.fromstring(xml_string):
        if "ListRecords" not in tag.tag:
             continue
        for record in tag:
             ....
 ```

Comment (ErnestaP) [reply dags/hindawi/utils.py]: I don't think that's the case. The idea of this is that the _record_ is in the ListRecords tag, and the only one way to check the name of the tag is by tag.tag. The data looks like this:
```xml
<ListRecords>
   <rocord>...</record>
   <rocord>...</record>
</ListRecords >
<somethingElse>....<somethingElse/>
```
So if I will take your suggestion, I will grab _somethingElse_ instead of _record_. And a _record_ I need

MJedr: yes, but you run your code when tag.tag is in `ListRecords`. You just continue to next tag if `ListRecords` in the `tag.tag`. It doesn't change anything in the code logic.

ErnestaP: aaa, I see now! **not in** , I skip the word **not**. Thank you for clarifying!

dags/hindawi/utils.py
ErnestaP: this condition is not in the loop, so continue is not valid here

dags/hindawi/hindawi.py
```diff
@@ -0,0 +1,46 @@
...
+    def save_xml_in_s3(dates: dict, repo: IRepository = HindawiRepository()):
+        parameters = HindawiParams(
+            from_date=dates["start_date"],
+            until_date=dates["until_date"],
+        ).get_params()
+        rest_api = HindawiApiClient(
+            base_url=os.getenv("HINDAWI_API_BASE_URL", "https://www.hindawi.com")
+        )
+        articles_metadata = rest_api.get_articles_metadata(parameters)
+        if articles_metadata:
```
ParthS007: If `articles_metadata` is not present, we can log to the user that since no metadata is received, nothing is uploaded to S3.

dags/hindawi/hindawi.py
```diff
@@ -0,0 +1,46 @@
...
+            return save_file_in_s3(data=articles_metadata, repo=repo)
```
ParthS007:
```suggestion
        if not articles_metadata:
            logging.warning("No new data is uploaded to s3")
            return
        return save_file_in_s3(data=articles_metadata, repo=repo)
```

dags/hindawi/hindawi_api_client.py
```diff
@@ -0,0 +1,37 @@
...
+
+class HindawiApiClient:
+    def __init__(
+        self,
+        base_url=None,
+        path_segments=None,
+    ):
+        self.base_url = base_url or getenv(
+            "HINDAWI_API_BASE_URL", "https://www.hindawi.com"
+        )
```
ParthS007, line 15:
```suggestion
        base_url="https://www.hindawi.com",
        path_segments=[],
    ):
        self.base_url = getenv("HINDAWI_API_BASE_URL",  base_url)
```

dags/hindawi/hindawi_api_client.py
```diff
@@ -0,0 +1,37 @@
...
+        self.path_segments = path_segments or ["oai-pmh", "oai.aspx"]
```
ParthS007, line 16:
```suggestion
        self.path_segments = path_segments if  path_segments else ["oai-pmh", "oai.aspx"]
```

dags/hindawi/hindawi_api_client.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def get_articles_metadata(self, parameters, doi=None):
+        all_path_segments = self.path_segments
+        if doi is not None:
+            all_path_segments = self.path_segments.append(doi)
```
ParthS007:
```suggestion
        if doi:
            self.path_segments.append(doi)
```

dags/hindawi/hindawi_api_client.py
```diff
@@ -0,0 +1,37 @@
...
+        request = Request(
+            base_url=self.base_url,
+            path_segments=all_path_segments,
```
ParthS007:
```suggestion
            path_segments=self.path_segments,
```

dags/hindawi/hindawi_api_client.py
```diff
@@ -0,0 +1,37 @@
...
+            parameters=parameters,
+        )
+        response_content = request.get_response_xml()
+        for tag in ElementTree.fromstring(response_content):
+            if "ListRecords" in tag.tag:
```
ParthS007, line 28: Adding a comment would be helpful why the loop break is required.

dags/hindawi/hindawi_file_processing.py
ParthS007: I read the previous comment. Maybe adding a `TODO` comment would be helpful to remember for coming PRs

dags/hindawi/hindawi_file_processing.py
```diff
@@ -0,0 +1,16 @@
...
+import airflow
+from airflow.decorators import dag, task
+
+
+@dag(start_date=airflow.utils.dates.days_ago(0))
+def hindawi_file_processing():
+    @task()
+    def parse(**kwargs):
+        if "params" in kwargs and "article" in kwargs["params"]:
+            article = json.loads(kwargs["params"]["article"])
```
ParthS007:
```suggestion
            article = json.loads(kwargs.get("params").get("article"))
```

dags/hindawi/utils.py
```diff
@@ -0,0 +1,48 @@
...
+                {
+                    "id": id,
+                    "record": ElementTree.tostring(record).decode("utf-8"),
+                }
+            )
+    return ids_and_records
+
+
+def trigger_file_processing_DAG(ids_and_records):
+    if len(ids_and_records) > 0:
```
ParthS007:
```suggestion
    if not ids_and_records:
       // Log that no ids or records to trigger the dag
       return
```

It will keep the loop out of the if and one less indent.

ParthS007 (commented):
Thanks for the PR, Left some suggestions.

dags/hindawi/hindawi_api_client.py
ErnestaP: from my perspective , my solution is better for me

dags/hindawi/hindawi_api_client.py
ErnestaP: from my perspective , my solution is better for me

dags/hindawi/utils.py
ErnestaP: we are already logging that there are no articles uploaded, I don't know is worth it to log it again

dags/common/utils.py
drjova: 👍

dags/hindawi/hindawi_api_client.py
```diff
@@ -0,0 +1,36 @@
...
+            path_segments=self.path_segments,
+            parameters=parameters,
+        )
+        response_content = request.get_response_xml()
+        for tag in ElementTree.fromstring(response_content):
+            if "ListRecords" in tag.tag:
+                break
+            return response_content
+
+    # TODO: don't forget to download pdf when it will be needed
```
drjova, line 32: left over?

dags/hindawi/hindawi_file_processing.py
drjova: 👍

dags/hindawi/hindawi_file_processing.py
```diff
@@ -0,0 +1,14 @@
+import airflow
+from airflow.decorators import dag, task
+
+
+@dag(start_date=airflow.utils.dates.days_ago(0))
+def hindawi_file_processing():
+    @task()
+    def parse(**kwargs):
+        if "params" in kwargs and "article" in kwargs["params"]:
+            record = kwargs.get("params").get("record")
```
drjova: this will fail you need something like `kwargs.get("params", {}).get("record")`

dags/hindawi/repository.py
```diff
@@ -0,0 +1,37 @@
...
+        if not objects:
+            return
+        dates = [obj.last_modified.strftime("%Y-%m-%d") for obj in objects]
+        return max(dates)
+
+    def save(self, key: str, obj: IO):
+        self.s3_bucket.upload_fileobj(obj, key)
+
+    def delete_all(self):
+        self.s3_bucket.objects.all().delete()
```
drjova, line 37: can we just inherit these methods? it seems that are the same with other publishers.

dags/hindawi/utils.py
```diff
@@ -0,0 +1,49 @@
...
+
+def save_file_in_s3(data, repo):
+    if data:
+        date_today = date.today().strftime("%Y-%m-%d")
+        prefix = datetime.now().strftime("%Y-%m-%dT%H:%M")
+        key = f"{date_today}/{prefix}.xml"
+        byte_file = BytesIO(data)
+        repo.save(key, byte_file)
+        return key
+
```
drjova: don't we have a similar function?

dags/hindawi/utils.py
```diff
@@ -0,0 +1,49 @@
...
+    for tag in ElementTree.fromstring(xml_string):
+        if "ListRecords" not in tag.tag:
+            continue
+        for record in tag:
+            if "record" not in record.tag:
+                continue
+            id = str(uuid.uuid4())
+            ids_and_records.append(
+                {
+                    "id": id,
```
drjova: why we generate a uuid?

dags/hindawi/utils.py
```diff
@@ -0,0 +1,49 @@
...
+def trigger_file_processing_DAG(ids_and_records):
+    if not ids_and_records:
+        return
+    for data in ids_and_records:
+            trigger_dag.trigger_dag(
+                dag_id="hindawi_file_processing",
+                run_id=data["id"],
+                conf={"record": data["record"]},
+                replace_microseconds=False,
+            )
```
drjova: it's something wrong with the indentation, aren't we using back and pre-commit?

requirements.txt
```diff
@@ -14,3 +14,4 @@ structlog==21.5.0
 bleach==4.1.0
 inspire-utils==3.0.25
 backoff==2.0.1
+uuid==1.30
```
drjova: we don't need it

drjova (commented):
Thank you so much @ErnestaP! I have added few comments 😄

requirements.txt
ErnestaP: why?

dags/hindawi/utils.py
ErnestaP: I just need an id for an object, what would be the right way to do it?

dags/hindawi/utils.py
ErnestaP: similar, but not the same, configuration is different

dags/hindawi/utils.py
drjova: but why do we need an uuid for the object?

dags/hindawi/utils.py
drjova: which part of configuration is different?

dags/hindawi/repository.py
ErnestaP: not with all, publishers, saving to the bucket is different. 
I can inherit  just find_the_last_uploaded_file_date, and  delete_all to more specific implementation

dags/hindawi/hindawi_api_client.py
ErnestaP: not really, reminder more. We don't need pdfs at this stage, but I can figure out form where are we getting them

dags/hindawi/hindawi_file_processing.py
ErnestaP: but if I'm already checking if the params are in kwargs, why do we need the default values?

dags/hindawi/utils.py
ErnestaP: I'm just using it as uniq string

dags/hindawi/utils.py
ErnestaP: key: sometimes it;s xml, sometimes it's json

dags/hindawi/hindawi_file_processing.py
drjova: you need a default value for `params` because if params is not in `kwargs` it will throw an exception.

dags/hindawi/repository.py
drjova: I see the point, it's ok to leave it like that now. Thanks!

dags/hindawi/utils.py
drjova: As we discussed IRL we don't need the id 😃

dags/hindawi/hindawi_file_processing.py
ErnestaP: but it will never reach this line below, if the 'prams' is not there, the whole function will return null

drjova: you don't need the `if` with `.get(params, {})`

dags/hindawi/utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+def save_file_in_s3(data, repo):
+    if data:
+        date_today = date.today().strftime("%Y-%m-%d")
+        prefix = datetime.now().strftime("%Y-%m-%dT%H:%M")
+        key = f"{date_today}/{prefix}.xml"
+        byte_file = BytesIO(data)
+        repo.save(key, byte_file)
+        return key
+
```
drjova:
```suggestion
    if not data:
        return None 
    date_today = date.today().strftime("%Y-%m-%d")
    prefix = datetime.now().strftime("%Y-%m-%dT%H:%M")
    key = f"{date_today}/{prefix}.xml"
    byte_file = BytesIO(data)
    repo.save(key, byte_file)
    return key

```

dags/common/utils.py
```diff
@@ -12,7 +12,12 @@ def set_harvesting_interval(repo, **kwargs):
...
+        or repo.find_the_last_uploaded_file_date()
+    )
```
```suggestion
    start_date = (
        kwargs.get('params', {}).get('start_date')
        or repo.find_the_last_uploaded_file_date()
    )
```

drjova (approved):

drjova:
@ErnestaP LGTM 😃 please squash!

PR #49, opened 2022-08-16
Hindawi parsing

ErnestaP:


dags/hindawi/hindawi_api_client.py
```diff
@@ -10,12 +10,12 @@ def __init__(
...
-        self.base_url = base_url or getenv(
-            "HINDAWI_API_BASE_URL", "https://www.hindawi.com"
-        )
-        self.path_segments = path_segments or ["oai-pmh", "oai.aspx"]
+        self.base_url = base_url
+        self.path_segments = path_segments
 
     def get_articles_metadata(self, parameters, doi=None):
+        self.base_url = getenv("HINDAWI_API_BASE_URL", "https://www.hindawi.com")
+        self.path_segments = ["oai-pmh", "oai.aspx"]
```
drjova: the `path_segments` are set in the init, why we want to override them here?

dags/hindawi/hindawi_file_processing.py
```diff
@@ -1,12 +1,58 @@
...
 
 @dag(start_date=airflow.utils.dates.days_ago(0))
 def hindawi_file_processing():
     @task()
     def parse(**kwargs):
-        return kwargs.get("params")
+        if "params" in kwargs and "record" in kwargs["params"]:
+            record = kwargs.get("params").get("record")
+            xml = ET.fromstring(record)
+            return parse_hindawi(xml)
```
drjova: Something like:
```suggestion
            record = kwargs.get("params", {}).get("record")
            if not record:
                return None
            xml = ET.fromstring(record)
            return parse_hindawi(xml)
```

dags/hindawi/hindawi_params.py
```diff
@@ -7,7 +7,7 @@ def __init__(
         from_date: str = (date.today() - timedelta(days=1)).strftime("%Y-%m-%d"),
         until_date: str = date.today().strftime("%Y-%m-%d"),
         verb: str = "listrecords",
-        set: str = "HINDAWI.AA",
+        set: str = "HINDAWI.AHEP",
```
drjova, line 10: are we getting only one set?

requirements.txt
```diff
@@ -14,3 +14,4 @@ structlog==21.5.0
 bleach==4.1.0
 inspire-utils==3.0.25
 backoff==2.0.1
+uuid==1.30
```
drjova: we don't need this

tests/units/hindawi/test_hindawi_parser.py
```diff
@@ -0,0 +1,133 @@
...
+        assert article["authors"] == authors
+
+
+def test_abstracts(parsed_articles):
+    for article in parsed_articles:
+        assert "abstract" in article
+        assert (
+            article["abstract"]
+            == "In recent years, the discovery in quarkonium spectrum of several states not predicted by the naive quark model has awakened a lot of interest. A possible description of such states requires the enlargement of the quark model by introducing quark-antiquark pair creation or continuum coupling effects. The unquenching of the quark models is a way to take these new components into account. In the spirit of the Cornell Model, this is usually done by coupling perturbatively a quark-antiquark state with definite quantum numbers to the meson-meson channel with the closest threshold. In this work we present a method to coupled quark-antiquark states with meson-meson channels, taking into account effectively the nonperturbative coupling to all quark-antiquark states with the same quantum numbers. The method will be applied to the study of the X(3872) resonance and a comparison with the perturbative calculation will be performed."
```
drjova: hmm pre-commit?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,200 @@
...
+            for affiliation in affiliations
+        ]
+        return parsed_affiliations
+
+    def _get_arxiv(self, article: ET.Element):
+        arxiv = article.find(
+            "ns0:metadata/ns1:record/ns0:datafield/[@tag='037']/ns0:subfield/[@code='9']",
+            self.prefixes,
+        )
+        if arxiv is None:
```
drjova:
```suggestion
        if not arxiv:
```

dags/hindawi/parser.py
```diff
@@ -0,0 +1,200 @@
...
+            self.logger.error("there is no Arxiv!")
```
drjova:
```suggestion
            self.logger.error("No arxiv id found.")
```

dags/hindawi/parser.py
```diff
@@ -0,0 +1,200 @@
...
+    def _get_affiliations(self, author):
+        affiliations = [
+            affiliation.text.replace("\n", "")
+            for affiliation in author.findall("ns0:subfield[@code='u']", self.prefixes)
+        ]
+        parsed_affiliations = [
+            {
+                "value": affiliation,
+                "organization": (",").join(affiliation.split(",")[:-1]),
+                "country": affiliation.split(", ")[-1:][0],
```
drjova: could you please give me an example?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,200 @@
...
+        )
+        if arxiv is None:
+            self.logger.error("there is no Arxiv!")
+            return None
+        if arxiv.text.lower() == "arxiv":
+            arxiv_value = article.find(
+                "ns0:metadata/ns1:record/ns0:datafield/[@tag='037']/ns0:subfield/[@code='a']",
+                self.prefixes,
+            ).text.lower()
+            return [{"value": ("").join(arxiv_value.replace("arxiv:", "").split())}]
```
drjova: hmm I don't understand this part of the code, why a tuple, and why hardcoded `arxiv:`?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,200 @@
...
+            article.findall(
+                "ns0:metadata/ns1:record/ns0:datafield/[@tag='700']", self.prefixes
+            )
+        )
+        authors = []
+        for author in authors_all:
+            author_ = {
+                "raw_name": author.find(
+                    "ns0:subfield[@code='a']", self.prefixes
+                ).text.replace("\n", ""),
```
drjova: do we really have `\n`?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,200 @@
...
+        return {
+            "copyright_statement": copyright_statement,
+            "copyright_year": copyright_year,
+        }
+
+    def _get_copyright_statement(self, article: ET.Element):
+        return self._get_copyright(article)["copyright_statement"]
+
+    def _get_copyright_year(self, article: ET.Element):
+        return self._get_copyright(article)["copyright_year"]
```
drjova: if we use it in such way, we run the `_get_copyright` function twice

dags/hindawi/parser.py
```diff
@@ -0,0 +1,200 @@
...
+    def _get_copyright(self, article: ET.Element):
+        copyright_statement = article.find(
+            "ns0:metadata/ns1:record/ns0:datafield/[@tag='542']/ns0:subfield/[@code='f']",
+            self.prefixes,
+        ).text
+        copyright_year = "".join(i for i in copyright_statement if i.isdigit())
+        return {
+            "copyright_statement": copyright_statement,
+            "copyright_year": copyright_year,
+        }
```
drjova: It would be efficient to do it with regex

dags/hindawi/parser.py
```diff
@@ -0,0 +1,200 @@
...
+    def _get_copyright_statement(self, article: ET.Element):
+        return self._get_copyright(article)["copyright_statement"]
+
+    def _get_copyright_year(self, article: ET.Element):
+        return self._get_copyright(article)["copyright_year"]
+
+    def _construct_license(self, url, license_type, version):
+        if url and license_type and version:
+            return {
+                "url": ("").join(url.split(" ")),
```
drjova: why `split` and `join`? do you have an example of a url that needs this?

dags/common/parsing/xml_extractors.py
```diff
@@ -11,15 +11,22 @@ def __init__(
...
         self.required = required
         self.default_value = default_value
         self.extra_function = extra_function
 
     def extract(self, article: ET.Element):
+        if self.prefixes:
+            node_with_prefix = self.extra_function(
+                article.find(self.source, self.prefixes).text
+            )
+            return node_with_prefix
```
drjova, line 29: why we need prefixes?

dags/hindawi/hindawi_api_client.py
ErnestaP: because for getting files, path segments are different 
url for getting metadata: `https://www.hindawi.com/oai-pmh/oai.aspx?verb=listrecords&set=HINDAWI.AHEP&metadataprefix=marc21&from=...`
url for getting files: `http://downloads.hindawi.com/journals/ahep/...`

dags/hindawi/hindawi_params.py
ErnestaP: I will ask Anne, I'm not sure, all possible sets:
    HINDAWI.AA (Advances in Astronomy)
    HINDAWI.AHEP (Advances in High Energy Physics)
    HINDAWI.AMP (Advances in Mathematical Physics)
    HINDAWI.JAS (Journal of Astrophysics)
    HINDAWI.JCMP (Journal of Computational Methods in Physics)
    HINDAWI.JGRAV (Journal of Gravity)

But from previous Anne's comment on Zulip, I believe it is just one:
`For Hindawi, we have only one journal, so the value for Journal name is anyway always the same.`

dags/hindawi/parser.py
ErnestaP: * DST Centre for Interdisciplinary Mathematical Sciences, Institute of Science, Banaras Hindu University, Varanasi 221005, India
* Department of Physics, Yantai University, Yantai 264005, China

dags/hindawi/parser.py
ErnestaP: Tuple? It's a dict
Because it might have `arxiv` in the value, but we don't need it. We need just digits. For example:
```xml
	<ns0:subfield code="a">arXiv:1608.04627</ns0:subfield>
```
we need just 1608.04627

dags/hindawi/parser.py
ErnestaP: I cannot do it, because arxiv Element object. For some reason:
`if not arxiv` returns `True`
and `if arxiv is None:` returns False

I understand that `not` should return True if the variable is not assigned, but in this case, we have the Element object assigned here
And `is None` when is None

dags/hindawi/parser.py
ErnestaP: actually, no

dags/hindawi/parser.py
ErnestaP: I checked more examples and I believe, my input file for the test was wrong

dags/common/parsing/xml_extractors.py
ErnestaP: we need prefixes because our marcxml fields have them, and on order to parse them:
```xml
<ns1:record ...
<ns0:datafield...
<ns0:subfield...
```
we need to pass prefixes to **find**
https://docs.python.org/3/library/xml.etree.elementtree.html#parsing-xml-with-namespaces

dags/common/parsing/parser.py
```diff
@@ -35,11 +35,19 @@ def __init__(self, extractors) -> None:
...
             if (value := extractor.extract(article)) is not None
         }
+        if "copyright" in extracted_value:
+            extracted_value["copyright_statement"] = extracted_value["copyright"][
+                "copyright_statement"
+            ]
+            extracted_value["copyright_year"] = extracted_value["copyright"][
+                "copyright_year"
+            ]
+        return extracted_value
```
drjova: why this logic is in common and not in Hindawi?

dags/hindawi/hindawi_api_client.py
```diff
@@ -10,12 +10,12 @@ def __init__(
...
     ):
-        self.base_url = base_url or getenv(
-            "HINDAWI_API_BASE_URL", "https://www.hindawi.com"
-        )
-        self.path_segments = path_segments or ["oai-pmh", "oai.aspx"]
+        self.base_url = base_url
+        self.path_segments = path_segments
 
     def get_articles_metadata(self, parameters, doi=None):
+        self.base_url = getenv("HINDAWI_API_BASE_URL", "https://www.hindawi.com")
```
drjova: why you overwriting here the `self.base_url`? shouldn't be this logic be in `__init__`?

dags/hindawi/hindawi_api_client.py
drjova: same, shouldn't be this logic be in `__init__`?

dags/hindawi/hindawi_api_client.py
```diff
@@ -29,7 +29,27 @@ def get_articles_metadata(self, parameters, doi=None):
                 break
             return response_content
 
-    # TODO: don't forget to download pdf when it will be needed
+    def get_file(self, parameters, extension, doi=None):
+        self.base_url = "http://downloads.hindawi.com"
+        self.path_segments = ["journals", "ahep"]
```
drjova: why overwriting the class variables?

dags/hindawi/hindawi_api_client.py
```diff
@@ -29,7 +29,27 @@ def get_articles_metadata(self, parameters, doi=None):
...
+
+        doi_part = doi.replace("10.1155/", "") + f".{extension}"
```
drjova: what if `doi` is None? it will break. Also why hardcoded `10.1155`?

dags/hindawi/hindawi_file_processing.py
```diff
@@ -1,12 +1,61 @@
...
+from common.enhancer import Enhancer
+from common.enricher import Enricher
+from hindawi.parser import HindawiParser
+from jsonschema import validate
+
+
+def parse_hindawi(xml):
+    parser = HindawiParser()
+    parsed = parser.parse(xml)
+    return parsed
```
drjova:
```suggestion
    parser = HindawiParser()
    return parser.parse(xml)
```

dags/hindawi/hindawi_file_processing.py
```diff
@@ -1,12 +1,61 @@
...
+
+
+def enhance_aps(parsed_file):
+    return Enhancer()("Hindawi", parsed_file)
```
drjova: is this `APS` or `Hindawi`?

dags/hindawi/hindawi_file_processing.py
```diff
@@ -1,12 +1,61 @@
...
+
+
+def enrich_aps(enhanced_file):
+    return Enricher()(enhanced_file)
```
drjova: is this `APS` or `Hindawi`?

dags/hindawi/hindawi_file_processing.py
```diff
@@ -1,12 +1,61 @@
...
+
+
+def aps_validate_record(enriched_file):
+    schema = requests.get(enriched_file["$schema"]).json()
+    validate(enriched_file, schema)
```
drjova: is this `APS` or `Hindawi`?

dags/hindawi/hindawi_file_processing.py
```diff
@@ -1,12 +1,61 @@
...
+    def enchance(parsed_file):
+        return parsed_file and enhance_aps(parsed_file)
+
+    @task()
+    def enrich(enhanced_file):
+        return enhanced_file and enrich_aps(enhanced_file)
+
+    @task()
+    def validate_record(enriched_file):
+        return enriched_file and aps_validate_record(enriched_file)
```
drjova: is this `APS` or `Hindawi`?

dags/hindawi/hindawi_params.py
drjova: We should double check that, it shouldn't be only one set here imo

dags/hindawi/hindawi_api_client.py
ErnestaP: https://github.com/SCOAP3/scoap3-next/blob/a9763de5886c64461ad901ad833941ada153e558/scoap3/modules/workflows/workflows/articles_upload.py#L302

dags/hindawi/parser.py
drjova: ☝️

dags/hindawi/parser.py
```diff
@@ -0,0 +1,212 @@
...
+                version = clean_url_parts.pop()
+                licencse_type = clean_url_parts.pop()
+                licenses.append(
+                    self._construct_license(
+                        license_url.text, licencse_type.upper(), version
+                    )
+                )
+            elif license_text.text:
+                licenses.append(license_text.text)
+        if len(licenses):
```
drjova: What this is checking? 🤔

requirements.txt
drjova: ☝️

dags/hindawi/parser.py
drjova: Why not regex? and why `("").join` we don't need the tuple

dags/hindawi/hindawi_api_client.py
drjova: We have to understand why this is there, could you please ask Anne?

dags/hindawi/parser.py
ErnestaP: if there any licenses, but this check doesn't really make sense, taking into account the whole code 😅

dags/hindawi/hindawi_api_client.py
drjova: This is still overwrites the class variable `self.path_segments`, why we have to do that?

dags/hindawi/hindawi_api_client.py
```diff
@@ -29,7 +32,30 @@ def get_articles_metadata(self, parameters, doi=None):
                 break
             return response_content
 
-    # TODO: don't forget to download pdf when it will be needed
+    def get_file(self, parameters, extension, doi=None):
+        self.path_segments_for_files = ["journals", "ahep"]
```
drjova: why we need a class variable for this?

dags/hindawi/hindawi_api_client.py
```diff
@@ -29,7 +32,30 @@ def get_articles_metadata(self, parameters, doi=None):
...
+        try:
+            doi_part = doi.replace("10.1155/", "") + f".{extension}"
+            request = Request(
+                base_url=self.files_url,
+                path_segments=self.path_segments_for_files + [doi_part],
+                parameters=parameters,
+            )
+            return request.get_response_bytes()
+        except AttributeError as e:
```
drjova: is this for `None` case? what about if the request fails?

dags/hindawi/parser.py
drjova: Let's have a better parsing for that, maybe regex would be more sufficient instead of splitting and having indexes :)

dags/hindawi/parser.py
drjova: ☝️

tests/units/hindawi/test_hindawi_parser.py
drjova: ☝️

dags/hindawi/hindawi_params.py
drjova, line 10: can we double check if this is the only set?

dags/hindawi/hindawi_api_client.py
```diff
@@ -2,25 +2,24 @@
...
-
+from requests.exceptions import RequestException
 
 class HindawiApiClient:
     def __init__(
         self,
-        base_url=None,
-        path_segments=None,
+        base_url=getenv("HINDAWI_API_BASE_URL", "https://www.hindawi.com"),
+        files_url=getenv("HINDAWI_API_FILES_URL", "http://downloads.hindawi.com") ,
```
drjova:
```suggestion
        base_url=None,
        files_url=None,
```

drjova: and overwrite them on __init__ :)

dags/hindawi/hindawi_api_client.py
```diff
@@ -29,8 +28,33 @@ def get_articles_metadata(self, parameters, doi=None):
                 break
             return response_content
 
-    # TODO: don't forget to download pdf when it will be needed
+    def get_file(self, parameters, extension, doi=None):
+        path_segments_for_files = ["journals", "ahep"]
+        try:
+            doi_part = doi.replace("10.1155/", "") + f".{extension}"
```
drjova, line 50: @ErnestaP could we double check with Anne why we need that?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,229 @@
...
+from common.parsing.extractor import IExtractor
+
+
+class HindawiParser(IParser):
+    prefixes = {
+            "marc": "http://www.loc.gov/MARC21/slim",
+            "ns0": "http://www.openarchives.org/OAI/2.0/",
+            "ns1": "http://www.loc.gov/MARC21/slim",
+            "xsi": "http://www.w3.org/2001/XMLSchema-instance",
+        }
```
drjova:
```suggestion
    prefixes = {
            "marc": "http://www.loc.gov/MARC21/slim",
            "ns0": "http://www.openarchives.org/OAI/2.0/",
            "ns1": "http://www.loc.gov/MARC21/slim",
            "xsi": "http://www.w3.org/2001/XMLSchema-instance",
    }
```


This should be fixed with pre-commit, have you install it? 😅

dags/hindawi/parser.py
drjova: @ErnestaP ☝️

dags/hindawi/hindawi_params.py
ErnestaP: yes, it's just one. I already talked with Anne about it

dags/hindawi/parser.py
ErnestaP: I was just removing whitespaces, strip() works here perfectly

dags/hindawi/parser.py
ErnestaP: I know, but I don't know the different solution. The last solution, didn't work, because I moved the logic to general parsing, to have this function used once

dags/hindawi/hindawi_params.py
drjova: This is the only set that we need.

dags/hindawi/hindawi_api_client.py
ErnestaP: Asked her, she will ask hindawi

dags/hindawi/hindawi_api_client.py
```diff
@@ -2,25 +2,29 @@
...
-        self.path_segments = path_segments or ["oai-pmh", "oai.aspx"]
+        self.files_url = files_url or getenv(
+            "HINDAWI_API_FILES_URL", "http://downloads.hindawi.com"
+        )
 
     def get_articles_metadata(self, parameters, doi=None):
+        path_segments = ["oai-pmh", "oai.aspx"]
         if doi:
-            self.path_segments.append(doi)
+            path_segments.append(doi)
```
drjova, line 27: do we have a cease with an article without doi? If the doi is None, what the request will return?

dags/hindawi/hindawi_api_client.py
```diff
@@ -29,8 +33,30 @@ def get_articles_metadata(self, parameters, doi=None):
...
+                parameters=parameters,
+            )
+            return request.get_response_bytes()
+        except AttributeError as e:
+            self.logger.error(f"No doi was passed for getting the file. Error: {e}")
+        except RequestException as e:
+            self.logger.error(f"Request failed witt exception: {e}")
+
+    def get_pdf_file(self, parameters, doi=None):
+        return self.get_file(parameters, "pdf", doi)
```
drjova: The responsibility logic of the pdf file's path should be in this function and not in the `get_file`. The same applies to `pdfa` and `xml`.

dags/hindawi/hindawi_api_client.py
```diff
@@ -29,8 +33,30 @@ def get_articles_metadata(self, parameters, doi=None):
...
+            request = Request(
+                base_url=self.files_url,
+                path_segments=path_segments_for_files + [doi_part],
+                parameters=parameters,
+            )
+            return request.get_response_bytes()
+        except AttributeError as e:
+            self.logger.error(f"No doi was passed for getting the file. Error: {e}")
+        except RequestException as e:
+            self.logger.error(f"Request failed witt exception: {e}")
```
drjova: the `self.logger` is not initialised anywhere, how come this is not caught on tests?

dags/hindawi/hindawi_api_client.py
ErnestaP: All articles by the date, not just one

dags/hindawi/hindawi_api_client.py
ErnestaP: because it comes from structlog.get_logger: https://github.com/cern-sis/workflows/blob/Hindawi-parsing/dags/hindawi/parser.py#L18

dags/hindawi/hindawi_api_client.py
drjova: so all the articles with dates don't have a doi?

dags/hindawi/hindawi_api_client.py
drjova: @ErnestaP this class doesn't have logger, only `HindawiParser` has it

dags/hindawi/parser.py
```diff
@@ -0,0 +1,201 @@
...
+                version = clean_url_parts.pop()
+                licencse_type = clean_url_parts.pop()
+                licenses.append(
+                    self._construct_license(
+                        license_url.text, licencse_type.upper(), version
+                    )
+                )
+            elif license_text.text:
+                licenses.append(license_text.text)
+            return licenses
```
drjova: don't we have another publisher with the same logic?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,201 @@
...
+        return copyright_statement
+
+    def _get_copyright_statement(self, article):
+        if self.statement is None:
+            self.statement = self._get_copyright(article)
+        return self.statement
+
+    def _get_copyright_year(self, article):
+        copyright_year = re.search(r"[0-9]{4}", self.statement).group(0)
+        return copyright_year
```
drjova: let's just add this logic to the parser class itself

dags/hindawi/utils.py
```diff
@@ -38,7 +38,7 @@ def trigger_file_processing_DAG(records):
     for record in records:
         trigger_dag.trigger_dag(
             dag_id="hindawi_file_processing",
-            conf=record,
+            conf={"record": record},
```
drjova, line 41: isn't this affecting all the other publishers? It's global change

dags/hindawi/parser.py
```diff
@@ -0,0 +1,201 @@
...
+            for affiliation in author.findall("ns0:subfield[@code='u']", self.prefixes)
+        ]
+        parsed_affiliations = [
+            {
+                "value": affiliation,
+                "organization": (",").join(affiliation.split(",")[:-1]),
+                "country": re.findall(r"[A-Z][a-z]*$", affiliation)[0],
+            }
+            for affiliation in affiliations
+        ]
```
drjova: this is prone to mistakes with index, could we please use regex?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,201 @@
...
+        )
+        if arxiv is None:
+            self.logger.error("No arxiv id_get_copyright_statement found.")
+            return None
+        if arxiv.text.lower() == "arxiv":
+            arxiv_value = article.find(
+                "ns0:metadata/ns1:record/ns0:datafield/[@tag='037']/ns0:subfield/[@code='a']",
+                self.prefixes,
+            ).text.lower()
+            return [{"value": arxiv_value.replace("arxiv:", "").strip()}]
```
drjova: This is not correct, we have ` arxiv.text.lower()` hence the `arxiv` might be `arXiv`. We have other similar cases, can we use a unified way to do it? i.e. regex?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def _get_journal_pages(self, article: ET.Element):
+        journal_pages = article.find(
+            "ns0:metadata/ns1:record/ns0:datafield/[@tag='773']/ns0:subfield/[@code='c']",
+            self.prefixes,
+        ).text
+        if "-" in journal_pages:
+            return journal_pages.split("-", 1)
+        else:
+            return journal_pages, ""
```
drjova:
```suggestion
        if "-" in journal_pages:
            return journal_pages.split("-", 1)

       return journal_pages, ""
```

dags/hindawi/parser.py
drjova: also, you are returning inside the `for` loop. Is it what we want?

dags/hindawi/parser.py
ErnestaP: yes, so all arxiv, arXiv, or even ARXIV will become arxiv, so it works with every variation. But yes, I can put regex

dags/hindawi/parser.py
ErnestaP: it is a regex, just findall returns array with one element in this case. re.search will return a match object from where we have to extract value with groups. Why findall doesn't work here?

dags/hindawi/utils.py
ErnestaP: no, this change is just for hindawi. You can notice it from the file path

dags/hindawi/parser.py
ErnestaP: similar, but not the same. Springer looks really similar, but actually, in hindawi, we're getting url and licenses statement, while in Springer we are getting license type and version, no statement

ErnestaP: and no, we don't want to return inside of for loop

dags/hindawi/parser.py
ErnestaP: how? 😅 I did it with a separate class in order to avoid calling the function twice. Otherwise, we have to parse marcxml  two times

dags/hindawi/hindawi_api_client.py
ErnestaP: Still waiting for reply

dags/common/parsing/generic_parsing.py
```diff
@@ -80,12 +80,14 @@ def parse_thesis_supervisors(value):
...
+        return article["publication_info"]
     return [
         {
-            "journal_title": article.get("journal_title", ""),
-            "journal_volume": article.get("journal_volume", ""),
-            "year": int(article.get("journal_year", 0)) or "",
-            "journal_issue": article.get("journal_issue", ""),
+            "journal_title": article.get("journal_title", [""]),
+            "journal_volume": article.get("journal_volume", [""]),
+            "year": int(article.get("journal_year", 0)) or [""],
```
drjova: why to default to a list? the schema it doesn't expect list

tests/units/hindawi/test_hindawi_parser.py
drjova: @ErnestaP could you please run pre-commit?

tests/units/hindawi/test_hindawi_parser.py
```diff
@@ -0,0 +1,167 @@
...
+def test_date_published(parsed_articles):
+    for article in parsed_articles:
+        assert "date_published" in article
+        assert article["date_published"] == "2019-05-02"
+
+
+def test_page_nr(parsed_articles):
+    for article in parsed_articles:
+        assert "page_nr" in article
+        assert article["page_nr"] == [7]
```
drjova: no need to assert both :) if `article['page_nr']` doesn't exist it will fail

tests/units/hindawi/test_hindawi_parser.py
```diff
@@ -0,0 +1,167 @@
...
+    with open(shared_datadir / "example2.xml") as file:
+        return ET.fromstring(file.read())
+
+
+@fixture()
+def parsed_article(hindawi_parser, article2):
+    return hindawi_parser._publisher_specific_parsing(article2)
+
+
+def test_title_(parsed_article):
```
drjova: could you please explain me this test?

tests/units/hindawi/test_hindawi_parser.py
ErnestaP: but if it exists, but the value is different?

dags/common/parsing/generic_parsing.py
drjova: @ErnestaP ☝️

tests/units/hindawi/test_hindawi_parser.py
```diff
@@ -0,0 +1,190 @@
...
+            id="test_license",
+        ),
+    ],
+)
+def test_hindawi_parsing(parsed_articles, expected, key):
+    for (
+        expected_value,
+        article,
+    ) in zip(expected, parsed_articles):
+        assert key in article
```
drjova: we don't have to do that :) just the next line it's enough

dags/hindawi/hindawi_api_client.py
ErnestaP: She said yes, it is for all of articles' pdfs, xmls, pdfas

dags/hindawi/hindawi_api_client.py
ErnestaP: We don't need, because we can make a request by:
* range of dates, we will get all articles published in this date range
* doi, just one specific article

tests/units/hindawi/test_hindawi_parser.py
ErnestaP: It doesn't fix the issue :(

dags/hindawi/hindawi_api_client.py
ErnestaP: Anne told, me that this situation is for all articles.
The first part of the doi reflects the publisher, which we are cutting off.
It is the way how Hindawi api works

tests/units/hindawi/cassettes/test_hindawi_get_file_with_wrong_DOI.yaml
```diff
@@ -0,0 +1,251 @@
...
+          - keep-alive
+        User-Agent:
+          - python-requests/2.27.1
+      method: GET
+      uri: https://downloads.hindawi.com/journals/ahep/2019%2F0000000.pdf?from=2022-01-01&until=2022-01-10&verb=listrecords&set=HINDAWI.AHEP&metadataprefix=marc21
+    response:
+      body:
+        string: '<?xml version="1.0" encoding="UTF-8"?>
+
+          <Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>JZHH81TFK875TGJ5</RequestId><HostId>94gSzuzTkNYyAy7eMoZe3COqIupiQxIp1Qh4ZekFnf7s4Hck1Meq7JDJ5vMZdXInCVenIbG/tsQ=</HostId></Error>'
```
drjova, line 145: @ErnestaP there is an error with hindawi requests!

PR #51, opened 2022-09-15
Iop harvesting

ErnestaP:


dags/common/sftp_service.py
```diff
@@ -48,4 +48,4 @@ def __connect(self) -> pysftp.Connection:
 
 class DirectoryNotFoundException(Exception):
     def __init__(self, *args: object) -> None:
-        super().__init__(*args)
+        super().__init__(*args)
```
ErnestaP: pre-commit

dags/iop/dag_pull_ftp.py
```diff
@@ -0,0 +1,41 @@
+import airflow
+import common.pull_ftp as pull_ftp
+from airflow.decorators import dag, task
+from common.repository import IRepository
+from common.sftp_service import SFTPService
+from iop.repository import IOPRepository
```
ErnestaP, line 6: separate third party imports and local ones

dags/common/cleanup.py
```diff
@@ -1,9 +1,13 @@
+import re
+
 from bleach import clean
 from bleach.html5lib_shim import Filter
 from bleach.sanitizer import Cleaner
 
 
 def clean_whitespace_characters(input):
+    if type(input) == list:
```
MJedr: I prefer to use `isinstance()`.
Also, is the list/string the only input possibility? Maybe you can add type annotation?

dags/common/cleanup.py
```diff
@@ -1,9 +1,13 @@
...
+        return input
     return " ".join(input.split())
```
MJedr, line 9: Won't `re.sub` or `str.replace` work for this case? IMO It would be more readable.

dags/common/cleanup.py
```diff
@@ -64,11 +68,9 @@ def remove_unnecessary_fields(obj):
 
 
 def remove_orcid_prefix(obj):
-    orcid_prefixes = ("orcid:", "https://orcid.org/", "http://orcid.org/")
+    pattern = re.compile(r"https{0,1}://orcid.org/|ORCID-|orcid:", flags=re.I)
```
MJedr: you added a flag `ignorecase` so maybe you can put the whole pattern in lowercase

dags/hindawi/hindawi_api_client.py
```diff
@@ -2,35 +2,62 @@
...
-            self.path_segments.append(doi)
+            path_segments.append(doi)
         request = Request(
             base_url=self.base_url,
-            path_segments=self.path_segments,
+            path_segments=path_segments,
             parameters=parameters,
         )
+        if "until" not in parameters:
+            raise AttributeError("until date is required")
```
MJedr: if it's required maybe it would be better to pass it as an arg or keyword arg (not in params)?

dags/hindawi/hindawi_api_client.py
```diff
@@ -2,35 +2,62 @@
...
+            doi_part = doi.replace("10.1155/", "") + f".{extension}"
+            request = Request(
+                base_url=self.files_url,
+                path_segments=path_segments_for_files + [doi_part],
+                parameters=parameters,
+            )
+            return request.get_response_bytes()
+        except RequestException:
+            # TODO return exception
+            self.logger.error("Request failed with exception.")
```
MJedr, line 58: maybe pass more data about the exception to logger

dags/hindawi/parser.py
```diff
@@ -0,0 +1,217 @@
...
+                source="ns0:metadata/ns1:record/ns0:datafield/[@tag='520']/ns0:subfield/[@code='a']",
+                prefixes=self.prefixes,
+                extra_function=lambda x: " ".join(x.split()),
+                default_value="",
+            ),
+            TextExtractor(
+                destination="title",
+                source="ns0:metadata/ns1:record/ns0:datafield/[@tag='245']/ns0:subfield/[@code='a']",
+                prefixes=self.prefixes,
+                extra_function=lambda x: x,
```
MJedr: it is the default, isn't it?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,217 @@
...
+        ) + article.findall(
+            "ns0:metadata/ns1:record/ns0:datafield/[@tag='700']", self.prefixes
+        )
+
+        authors = []
+        for author in authors_all:
+            raw_name = author.find("ns0:subfield[@code='a']", self.prefixes).text
+            if not raw_name:
+                continue
+            author_ = {
```
MJedr: maybe a better name?

dags/hindawi/parser.py
```diff
@@ -0,0 +1,217 @@
...
+                source="ns0:metadata/ns1:record/ns0:datafield/[@tag='245']/ns0:subfield/[@code='a']",
+                prefixes=self.prefixes,
+                extra_function=lambda x: x,
+                default_value="",
+            ),
+            TextExtractor(
+                destination="date_published",
+                source="ns0:metadata/ns1:record/ns0:datafield/[@tag='260']/ns0:subfield/[@code='c']",
+                prefixes=self.prefixes,
+                extra_function=lambda x: x,
```
MJedr: same

dags/hindawi/xml_extractors.py
```diff
@@ -0,0 +1,71 @@
...
+        if value:
+            return value
+        return ""
+
+    def _get_extracted_value(self, node):
+        if self.required and node is None:
+            raise RequiredFieldNotFoundExtractionError(self.source)
+        if node is None:
+            return self.default_value
+        pattern = re.compile(r"[\n\t]*")
```
MJedr: maybe you can add a better naming what is this pattern

dags/hindawi/xml_extractors.py
```diff
@@ -0,0 +1,71 @@
...
+            for value in values:
+                extracted_values.append(
+                    "".join(
+                        [pattern.sub("", value.text)]
+                        + [
+                            pattern.sub("", ET.tostring(el).decode("UTF-8"))
+                            for el in value
+                        ]
+                    )
+                )
```
MJedr: maybe split it in more statements?

dags/hindawi/xml_extractors.py
```diff
@@ -0,0 +1,71 @@
...
+            print(extracted_values)
```
MJedr: maybe use logger instead?

dags/iop/repository.py
```diff
@@ -0,0 +1,62 @@
...
+        ret_dict = {}
+        filenames = []
+        filenames = (
+            filenames_to_process
+            if filenames_to_process
+            else self.__find_all_extracted_files()
+        )
+        for file in filenames:
+            file_parts = file.split("/")
+            last_part = file_parts[-1]
```
MJedr: you can use `os.path.basename` to get the last part

dags/iop/repository.py
```diff
@@ -0,0 +1,62 @@
...
+            filename_without_extension = last_part.split(".")[0]
+            if filename_without_extension not in ret_dict.keys():
```
MJedr: `if filename_without_extension not in ret_dict` checks if it's in keys

dags/iop/repository.py
```diff
@@ -0,0 +1,62 @@
...
+                ret_dict[filename_without_extension] = dict()
```
MJedr: you can use `defaultdict(dict)` instead dict for `ret_dict`, you won't need to create a new dict here

dags/iop/repository.py
```diff
@@ -0,0 +1,62 @@
...
+            ret_dict[filename_without_extension][
+                "xml" if self.is_meta(last_part) else "pdf"
```
MJedr: maybe extract this before, assign a meaningful variable name and then refer to this value, so 
```
extension = "xml" if self.is_meta(last_part) else "pdf"
ret_dict[filename_without_extension][extension] = file
```

MJedr (commented):
few suggestions from me :)

dags/hindawi/hindawi_api_client.py
ErnestaP: I have just realized that code-wise, the situation should never happen because the until date, if it's not given, has to be set to current date

dags/hindawi/hindawi_api_client.py
ErnestaP: This fix has to go to another branch

dags/hindawi/hindawi_api_client.py
ErnestaP: This fix has to go to another branch

dags/hindawi/xml_extractors.py
ErnestaP: it's removed completely

dags/hindawi/parser.py
ErnestaP: no, just extra functionality for the text parser, if the value has to be modified. In this case nothing is done with it

dags/hindawi/parser.py
ErnestaP: This fix has to go to another branch, because this PR is for IOP, not hindawi publisher

dags/hindawi/xml_extractors.py
ErnestaP: This fix has to go to another branch, because this PR is for IOP, not hindawi publisher

dags/common/cleanup.py
ErnestaP: in my opinion, regex might be a bit difficult, because we need to catch:
 * new lines
 * tabs
 * double spaces
 * trailing spaces
I can find a different solutions but might become hard to read

PR #53, opened 2022-09-28
OUP harvesting

ErnestaP:


dags/oup/oup_pull_ftp.py
```diff
@@ -0,0 +1,37 @@
+import airflow
+import common.pull_ftp as pull_ftp
+from airflow.decorators import dag, task
```
ErnestaP, line 3: Separate imports

tests/integration/oup/test_oup_pull_ftp.py
```diff
@@ -0,0 +1,58 @@
+import datetime
+
+import pytest
+from airflow import DAG
```
ErnestaP, line 4: imports

tests/units/oup/test_oup_repository.py
```diff
@@ -0,0 +1,73 @@
...
+
+def test_find_all(boto3_fixture):
+    repo = OUPRepository()
+    assert repo.find_all() == FIND_ALL_EXPECTED_VALUES
+
+
+def test_find_all_extracted_files(boto3_fixture):
+    repo = OUPRepository()
+    assert (
+        repo._OUPRepository__find_all_extracted_files()
```
drjova: I'm very confused, what is this?

tests/integration/oup/test_oup_sftp.py
```diff
@@ -0,0 +1,24 @@
...
+    yield SFTPService(dir="upload/oup")
+
+
+def test_list_oup_files(oup_client_fixture: SFTPService):
+    data_files = os.listdir(pathlib.Path().resolve().__str__() + "/data/oup")
+    assert len(oup_client_fixture.list_files()) == len(data_files)
+
+
+def test_get_oup_file(oup_client_fixture: SFTPService):
+    data_path = pathlib.Path().resolve().__str__() + "/data/oup"
```
drjova: why we need this? we have `pytest-datadir`

tests/units/oup/test_oup_repository.py
ErnestaP: it is testing, does the [__find_all_extracted_files](https://github.com/cern-sis/workflows/blob/oup/dags/oup/repository.py#L52) method works as expected, it means, does it returns just XML files and pdfs

tests/integration/oup/test_oup_sftp.py
ErnestaP: This test does not make sense in the end. Because it is to check does the local ftp server has the same amount of files or not, as the data dir (from data dir, files are uploaded to local ftp)

tests/integration/oup/test_oup_pull_ftp.py
```diff
@@ -0,0 +1,58 @@
...
+        OUPSFTPService(),
+        repo,
+        get_logger().bind(class_name="test_logger"),
+        **{"params": {}},
+    )
+    print(repo.find_all())
+    assert len(repo.find_all()) == 3
+
+
+def __test_dagrun_state(dagrun: DagRun):
```
drjova: why `__`?

tests/units/oup/test_oup_repository.py
drjova: but this is a private method, it should be test it the same way we are using it i.e. the `find_all()`

tests/integration/oup/test_oup_pull_ftp.py
```diff
@@ -0,0 +1,58 @@
...
+    repo = OUPRepository()
+    repo.delete_all()
+    assert len(repo.find_all()) == 0
+    migrate_from_ftp(
+        OUPSFTPService(),
+        repo,
+        get_logger().bind(class_name="test_logger"),
+        **{"params": {}},
+    )
+    print(repo.find_all())
```
drjova: leftover

tests/integration/oup/test_oup_pull_ftp.py
ErnestaP: because it was an additional function that was used in another test. Renamed it

tests/units/oup/test_oup_repository.py
ErnestaP: But in find_all, we are getting an array of dicts with xml and pdf. And I want to check is the method gives me just xmls. The test for find_all is here:https://github.com/cern-sis/workflows/blob/fa015c884473cfb169998a172ca810e73e35145a/tests/units/oup/test_oup_repository.py#L46-L48

Or you had in mind something else?

ErnestaP: the test for find_all is enough?

dags/oup/oup_process_file.py
```diff
@@ -0,0 +1,9 @@
+import airflow
+from airflow.decorators import dag
+
+
+@dag(schedule_interval=None, start_date=airflow.utils.dates.days_ago(0))
+def oup_process_file():
+    pass
```
ErnestaP, line 7: Will be implemented on the other PR. We need to have this DAG because of the trigger_file_processing task in harvesting DAG. trigger_file_processing is the last task in harvesting DAG, the task passes files for parsing
https://github.com/cern-sis/workflows/blob/d3f99b5c294a81a64b7d96c1841dec6ddb067fc5/dags/oup/oup_pull_ftp.py#L42-L51

dags/oup/repository.py
```diff
@@ -0,0 +1,60 @@
...
+        self.s3 = S3Service(os.getenv("OUP_BUCKET_NAME", "oup"))
+
+    def get_all_raw_filenames(self):
+        return [
+            os.path.basename(f.key)
+            for f in self.s3.objects.filter(Prefix=self.ZIPED_DIR).all()
+        ]
+
+    def find_all(self, filenames_to_process=None):
+        ret_dict = {}
```
MJedr: what is `ret`? can you add a more verbose names? :D

dags/oup/repository.py
```diff
@@ -0,0 +1,60 @@
...
+        filenames = []
+        filenames = (
```
MJedr: why you redefine it? (Or maybe why you have to define `filenames=[]`

dags/oup/repository.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+        for file in filenames:
+            last_part = os.path.basename(file)
+            filename_without_extension = last_part.split(".")[0]
+            ret_dict.setdefault(filename_without_extension, defaultdict(dict))
+            extension = "xml" if self.is_meta(last_part) else "pdf"
+            ret_dict[filename_without_extension].setdefault(extension, file)
+        return list(ret_dict.values())
+
+    def find_by_id(self, id: str):
```
MJedr: maybe `get_by_id` since you are downloading it?

ErnestaP: makes sense, this change will cause 16 files change :D

dags/oup/repository.py
```diff
@@ -0,0 +1,61 @@
...
+            filenames_to_process
+            if filenames_to_process
+            else self.__find_all_extracted_files()
+        )
+        if not filenames:
+            return []
+        for file in filenames:
+            last_part = os.path.basename(file)
+            filename_without_extension = last_part.split(".")[0]
+            grouped_files.setdefault(filename_without_extension, defaultdict(dict))
```
MJedr: one last thing, the comment was lost somewhere: do you really need a default dict there? It is a very good structure but takes at least twice of the memory as a standard `dict` does. In this use case I think you can just set a value of a dict as a simple dict `{value: key}`, or am I wrong?

ErnestaP: submitted the change :)

ErnestaP: it requires if/else and check-ups, I am not sure if it's better solution

tests/units/oup/test_oup_repository.py
```diff
@@ -38,6 +38,7 @@ def boto3_fixture():
 
 def test_find_all(boto3_fixture):
     repo = OUPRepository()
+    print(repo.find_all())
```
MJedr: leftover

MJedr (approved):

PR #56, opened 2022-10-10
small fix for text extraction

ErnestaP:
Running on QA one article crashed with ''error: _get_extracted_value
    [pattern.sub("", node.text) or ""]
TypeError: expected string or bytes-like object''

Small fix provided


dags/hindawi/xml_extractors.py
```diff
@@ -20,7 +20,7 @@ def _get_extracted_value(self, node):
             return self.default_value
         pattern = re.compile(r"[\n\t]*")
         extracted_value = "".join(
-            [pattern.sub("", node.text) or ""]
+            [node.text and pattern.sub("", node.text) or ""]
```
drjova: could you please add a test? Also it's getting too complex, could you please separate the logic?

drjova: @ErnestaP ☝️

tests/units/hindawi/test_hindawi_text_extractor.py
```diff
@@ -0,0 +1,56 @@
...
+        source="ns0:metadata/ns1:record/ns0:datafield/[@tag='024']/ns0:subfield/[@code='a']",
+        destination="dois",
+        prefixes=prefixes,
+    )
+
+
+@fixture()
+def articles(shared_datadir):
+    articles = []
+    files = ["example1.xml", "example2.xml"]
```
drjova: do they have the case that we want to test?

drjova: @ErnestaP ☝️

ErnestaP: yes, the titles and abstracts have a lot of inner html's, like math expressions, italics

drjova: @ErnestaP  hmm why it didn't fail before in the tests?

ErnestaP: because I didn't parse with mathlib at all, we left it for solving in the future.
The goal before was to extract the first string, to check are we parsing the right part in xml

ErnestaP: and also it was different parsing function,which was taking just the first string, but not the whole value

dags/hindawi/xml_extractors.py
```diff
@@ -18,12 +18,13 @@ def _get_extracted_value(self, node):
...
-        pattern = re.compile(r"[\n\t]*")
+        pattern = re.compile(r"[\n\t]{0,}" + r"\s{2,}")
         extracted_value = "".join(
-            [pattern.sub("", node.text) or ""]
-            + [pattern.sub("", ET.tostring(el).decode("ascii")) for el in node]
+            [node.text or ""] + [ET.tostring(el).decode("ascii") for el in node]
         )
-        return self.extra_function(extracted_value)
+        final_value = self.extra_function(pattern.sub("", extracted_value))
+        self.logger.info(f"Extracted {self.destination}={final_value}")
```
drjova:
```suggestion
        self.logger.info(f"Extracted value", field={self.desigtation}, value={final_value})
```

tests/units/hindawi/test_hindawi_text_extractor.py
```diff
@@ -0,0 +1,56 @@
...
+    return HindawiTextExtractor(
+        source="ns0:metadata/ns1:record/ns0:datafield/[@tag='024']/ns0:subfield/[@code='nothing']",
+        destination="nothing",
+        prefixes=prefixes,
+    )
+
+
+def test_parsed_articles_with_error(hindawi_not_excitant_field_extractor, articles):
+    with raises(RequiredFieldNotFoundExtractionError):
+        [hindawi_not_excitant_field_extractor.extract(article) for article in articles]
```
drjova: unfortunately, this test is testing just the test logic and not the real logic :) In order to make sure the code works properly we have to add a xml file with missing field

dags/hindawi/parser.py
```diff
@@ -54,6 +55,7 @@ def __init__(self) -> None:
             TextExtractor(
                 destination="page_nr",
                 source="ns0:metadata/ns1:record/ns0:datafield/[@tag='300']/ns0:subfield/[@code='a']",
+                required=False,
```
drjova, line 58: by default is `False` no need to specify it

dags/hindawi/xml_extractors.py
```diff
@@ -18,12 +18,15 @@ def _get_extracted_value(self, node):
...
+        pattern = re.compile(r"[\n\t]{0,}" + r"\s{2,}")
         extracted_value = "".join(
-            [pattern.sub("", node.text) or ""]
-            + [pattern.sub("", ET.tostring(el).decode("ascii")) for el in node]
+            [node.text or ""] + [ET.tostring(el).decode("ascii") for el in node]
         )
-        return self.extra_function(extracted_value)
+        final_value = self.extra_function(pattern.sub("", extracted_value))
+        self.logger.info(
+            "Extracted value field={self.destination}, value={final_value}"
```
drjova:
```suggestion
            "Extracted value", field={self.destination}, value={final_value}"
```

drjova: https://www.structlog.org/en/stable/why.html#easier-logging

dags/hindawi/parser.py
ErnestaP: No, by default is True: https://github.com/cern-sis/workflows/blob/hindawi_parser_fix/dags/common/parsing/xml_extractors.py#L12

dags/hindawi/xml_extractors.py
ErnestaP: A, I see. I think it has to be without curly brackets

tests/units/hindawi/test_hindawi_parser.py
```diff
@@ -187,3 +244,60 @@ def test_hindawi_parsing(parsed_articles, expected, key):
...
+        # DOI is required, it has to crash in the beginning
+        if file_name_with_missing_data == "without_doi.xml":
+            with pytest.raises(RequiredFieldNotFoundExtractionError):
+                hindawi_parser._publisher_specific_parsing(
+                    articles_with_missing_data[file_name_with_missing_data]
+                )
+        else:
+            hindawi_parser._publisher_specific_parsing(
+                articles_with_missing_data[file_name_with_missing_data]
+            )
```
drjova: it's better not to add additional logic to the tests, it would be better to create two separate tests. Also the `else` doesn't test anything :)

tests/units/hindawi/test_hindawi_parser.py
```diff
@@ -187,3 +244,60 @@ def test_hindawi_parsing(parsed_articles, expected, key):
...
+
+
+@pytest.fixture()
+def parsed_article_without_page_nr(hindawi_parser, articles_with_missing_data):
+    for file_name_with_missing_data in articles_with_missing_data.keys():
+        if file_name_with_missing_data == "without_page_nr.xml":
+            return hindawi_parser._publisher_specific_parsing(
+                articles_with_missing_data[file_name_with_missing_data]
+            )
```
drjova: same, it's better not to add extra logic to the tests

tests/units/hindawi/test_hindawi_parser.py
ErnestaP: it does because this test function doesn't have the return sentence. So the crashing one still we be executed on the line 269

drjova: Then we only need to test the case that raises `RequiredFieldNotFoundExtractionError`

ErnestaP: yes, I agree. The idea was to see the difference between two fields: required and not required. One of them, required one, has to raise an error while another one - not

tests/units/hindawi/test_hindawi_parser.py
```diff
@@ -187,3 +244,52 @@ def test_hindawi_parsing(parsed_articles, expected, key):
...
+    return enhance_hindawi(parsed_generic_article)
+
+
+@pytest.fixture
+def enrich_article(enhance_article):
+    return enrich_hindawi(enhance_article)
+
+
+def test_hindawi_validated_record(enrich_article):
+    hindawi_validate_record(enrich_article)
```
drjova: We don't need all these fixtures, since are only used in 1 test

tests/units/hindawi/test_hindawi_parser.py
```diff
@@ -187,3 +244,52 @@ def test_hindawi_parsing(parsed_articles, expected, key):
...
+def enhance_article(parsed_generic_article):
+    return enhance_hindawi(parsed_generic_article)
+
+
+@pytest.fixture
+def enrich_article(enhance_article):
+    return enrich_hindawi(enhance_article)
+
+
+def test_hindawi_validated_record(enrich_article):
```
drjova:
```suggestion
def test_hindawi_record_validation(enrich_article):
```

drjova (approved):
@ErnestaP please squash the commits

PR #57, opened 2022-10-31
workflows: IOP parser fix

ErnestaP:
* Added a check while extracting text from Elementree object.
* Parser won't try to extract text from Elemntree obj, which are None (for example: arxiv_id).
* FIX: before extracting text on Custom Extractor, the value of Elementree object is checked.

### DIFF
--- a/dags/common/parsing/xml_extractors.py
+++ b/dags/common/parsing/xml_extractors.py
@@ -56,9 +56,9 @@ class AttributeExtractor(IExtractor):
         self.required = required
 
     def extract(self, article: ET.Element):
-        value = self.extra_function(article.find(self.source).get(self.attribute))
-        if value:
-            return value
+        value = article.find(self.source)
+        if value is not None:
+            return self.extra_function(value.get(self.attribute))
         if self.required and value is None:
             raise RequiredFieldNotFoundExtractionError(self.source)
         return self.default_value
--- a/dags/iop/parser.py
+++ b/dags/iop/parser.py
@@ -1,0 +1,254 @@
+import datetime
+import re
+import xml.etree.ElementTree as ET
+
+from common.parsing.parser import IParser
+from common.parsing.xml_extractors import AttributeExtractor, CustomExtractor
+from hindawi.xml_extractors import HindawiTextExtractor as TextExtractor
+from structlog import get_logger
+
+# FIXME used Hindawi xml extractor, because it has mathlib string handling
+
+
+class IOPParser(IParser):
+    def __init__(self) -> None:
+        self.logger = get_logger().bind(class_name=type(self).__name__)
+        self.article_type_mapping = {
+            "research-article": "article",
+            "corrected-article": "article",
+            "original-article": "article",
+            "correction": "corrigendum",
+            "addendum": "addendum",
+            "editorial": "editorial",
+        }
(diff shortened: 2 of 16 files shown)


dags/common/parsing/xml_extractors.py
```diff
@@ -76,7 +76,7 @@ def __init__(
 
     def extract(self, article: ET.Element):
         value = self.extraction_function(article)
-        if value:
+        if value is not None:
             return value
```
ParthS007: This would be enough IMO.
```suggestion
        if value:
            return value
```

dags/iop/parser.py
```diff
@@ -140,9 +141,10 @@ def _get_arxiv_eprints(self, article):
         arxiv_eprints = []
         arxivs_value = article.find(
             "front/article-meta/custom-meta-group/custom-meta/meta-value"
-        ).text.lower()
-        pattern = re.compile(r"(arxiv:|v[0-9]$)", flags=re.I)
-        arxiv_eprints.append({"value": pattern.sub("", arxivs_value)})
+        )
+        if arxivs_value is not None:
```
ParthS007:
```suggestion
        if arxivs_value:
```

dags/iop/parser.py
```diff
@@ -239,16 +241,21 @@ def _get_license(self, article: ET.Element):
...
-                "journal_title": article.find(
-                    "front/journal-meta/journal-title-group/journal-title",
-                ).text,
-                "journal_volume": article.find(
-                    "front/article-meta/volume",
-                ).text,
-                "journal_year": self._get_journal_year(article),
-                "journal_issue": article.find("front/article-meta/issue").text,
-                "journal_artid": article.find("front/article-meta/elocation-id").text,
+                "journal_title": journal_title is not None and journal_title.text,
```
ParthS007: Similar for below ones:

```suggestion
                "journal_title": journal_title.text if journal_title else None,
```

ParthS007 (commented):
Some comments!

dags/common/parsing/xml_extractors.py
pamfilos: https://stackoverflow.com/a/7816439

ErnestaP: @ParthS007 , sadly is not enough 
Screen Shot 2022-10-27 at 14 11 26

dags/iop/parser.py
ErnestaP: same as before

dags/iop/parser.py
ErnestaP: same as before

dags/iop/parser.py
```diff
@@ -125,9 +125,10 @@ def _get_journal_doctype(self, article: ET.Element):
         journal_doctype = self.article_type_mapping[raw_journal_doctype]
         if "other" in journal_doctype:
             doi = article.find("front/article-meta/article-id/[@pub-id-type='doi']")
-            self.logger.msg(
-                f"There are unmapped article types for article {doi} with types {raw_journal_doctype}"
-            )
+            if doi:
```
```suggestion
            if doi is not None:
```
?

ErnestaP: yes! Thank you

PR #63, opened 2022-11-02
workflows: IOP parser fix

ErnestaP:
* Added a check while extracting text from Elementree object.
* Parser won't try to extract text from Elemntree obj, which are None (for example: arxiv_id).
* FIX: before extracting text on Custom Extractor, the value of Elementree object is checked.
* Parser assigns collaboration value to the wrong field. It creates it as author and value put in given_name field
* It happens because the collaborations are in the author's section in xml file.
* FIX: created a new extractor for collaborationsm which adds values that have collaboration word in the string.
* If the author has one or more affiliations, just the first one is added to the final JSON.
* FIX: taken all affiliation values by ref-id, which belong to the author.
* ref: https://github.com/cern-sis/issues-scoap3/issues/98"


ErnestaP:
Sorry, had to reopen the brach again, because I accidentally removed all changes and closed the branch

dags/common/utils.py
```diff
@@ -29,3 +30,7 @@ def construct_license(url, license_type, version):
     logger.error(
         "License is not given, or missing arguments.",
     )
+
+def get_text_value(element:  ET.Element):
+    if element is not None:
+        return element.text
```
drjova: wrong end of the file

tests/units/iop/test_iop_parser.py
```diff
@@ -43,19 +58,30 @@ def test_iop_record_validation(iop_parser, parsed_article_without_page_nr):
...
         with open(shared_datadir / file) as f:
             articles.append(ET.fromstring(f.read()))
-    return articles
+    yield articles
 
 
 @fixture
 def parsed_articles(iop_parser, articles):
-    return [iop_parser._publisher_specific_parsing(article) for article in articles]
+    yield [iop_parser._publisher_specific_parsing(article) for article in articles]
```
drjova, line 538:
```suggestion
     for article in articles:
         yield iop_parser._publisher_specific_parsing(article)
```

ErnestaP: why it is better?

ErnestaP: https://stackoverflow.com/questions/59536684/pytest-fixture-yield-list-of-dict-throws-yield-fixture-function-has-more-than-o
...But fixture functions can only yield exactly one value....

dags/iop/parser.py
```diff
@@ -170,27 +189,30 @@ def _get_affiliations(self, article):
...
         return authors
 
     def _get_date(self, date):
         return datetime.date(
-            day=int(date.find("day").text),
-            month=int(date.find("month").text),
-            year=int(date.find("year").text),
+            day=int(get_text_value(date.find("day"))),
+            month=int(get_text_value(date.find("month"))),
+            year=int(get_text_value(date.find("year"))),
```
drjova: `get_text_value` if the `key` doesn't exist it will return `None`. Hence the `int` will fail. For these type of cases we should have tests :)

tests/units/iop/test_iop_parser.py
```diff
@@ -398,7 +515,10 @@ def test_affiliations(parsed_articles):
             },
         ],
     ]
+
     for author, article in zip(authors, parsed_articles):
+        print(article["authors"])
+        print("\n")
```
drjova: leftovers

tests/units/iop/test_iop_parser.py
```diff
@@ -471,3 +591,9 @@ def test_get_license(parsed_articles):
     ]
     for license, article in zip(licenses, parsed_articles):
         assert article["license"] == license
+
+
+def test_get_collaborations(parsed_articles):
```
drjova: This comment also is related for the other tests. Why we are adding `parsed_articles` fixture since we have to iterate. We can use the the `params` from `pytest` to have our code [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) and without extra logic

dags/iop/parser.py
ErnestaP: Good point, I will write a test when all not required fields are missing

PR #65, opened 2022-11-29
Iop parser doi

ErnestaP:
IOP doi parsing + tests


tests/units/iop/test_parser/test_dois.py
```diff
@@ -0,0 +1,31 @@
...
+from pytest import fixture, raises
+
+
+@fixture
+def parser():
+    return IOPParser()
+
+
+def test_happy_path(shared_datadir, parser):
+    with open(shared_datadir / "just_required_fields.xml") as file:
```
drjova: please check the documentation how to read the files without using `open` https://pypi.org/project/pytest-datadir/

tests/units/iop/test_parser/test_dois.py
```diff
@@ -0,0 +1,31 @@
...
+from iop.xml_extractors import RequiredFieldNotFoundExtractionError
+from pytest import fixture, raises
+
+
+@fixture
+def parser():
+    return IOPParser()
+
+
+def test_happy_path(shared_datadir, parser):
```
drjova: `test_dois`

dags/iop/parser.py
```diff
@@ -0,0 +1,32 @@
...
+    def __init__(self) -> None:
+        self.logger = get_logger().bind(class_name=type(self).__name__)
+        self.article_type_mapping = {
+            "research-article": "article",
+            "corrected-article": "article",
+            "original-article": "article",
+            "correction": "corrigendum",
+            "addendum": "addendum",
+            "editorial": "editorial",
+        }
```
drjova: we can move them as class attribute, since are static

dags/iop/parser.py
```diff
@@ -0,0 +1,32 @@
...
+                extraction_function=self._get_dois,
+                required=True,
+            ),
+        ]
+        super().__init__(extractors)
+
+    def _get_dois(self, article: ET.Element):
+        dois_raw = article.find("front/article-meta/article-id/[@pub-id-type='doi']")
+        if dois_raw is not None:
+            self.logger.msg(f"Starting to parse article with doi {dois_raw.text}")
```
```python
dois = dois_raw.text
self.logger.msg(f"Parsing dois for article", dois= dois)
return dois

```

dags/iop/xml_extractors.py
```diff
@@ -0,0 +1,112 @@
...
+        elif not self.required and value:
+            return value
+        elif self.required and value is None:
+            raise RequiredFieldNotFoundExtractionError(self.destination)
+        elif not bool(self.default_value) and self.required:
+            raise RequiredFieldNotFoundExtractionError(self.destination)
+        return self.default_value
+
+
+class CustomExtractor(IExtractor):
```
drjova: the code of these classes is very similar, also I will add them in common.

dags/iop/xml_extractors.py
```diff
@@ -0,0 +1,112 @@
...
+        elif not self.required and value:
+            return value
+        elif self.required and value is None:
+            raise RequiredFieldNotFoundExtractionError(self.destination)
+        elif not bool(self.default_value) and self.required:
+            raise RequiredFieldNotFoundExtractionError(self.destination)
+        return self.default_value
+
+
+class ConstantExtractor(IExtractor):
```
drjova: where is this used?

dags/iop/xml_extractors.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(self, destination, constant) -> None:
+        super().__init__(destination)
+        self.constant = constant
+
+    def extract(self, _: ET.Element):
+        return self.constant
+
+
+class RequiredFieldNotFoundExtractionError(RuntimeError):
```
drjova: why `RuntimeError` error? https://docs.python.org/3/library/exceptions.html?highlight=runtimeerror#RuntimeError This case is a very specific error

dags/iop/xml_extractors.py
```diff
@@ -0,0 +1,112 @@
...
+        if bool(node) and all(node) and self.required:
+            return self.extra_function(article.find(self.source).text)
+        elif not self.required and node:
+            return self.extra_function(article.find(self.source).text)
+        elif self.required and node is None:
+            raise RequiredFieldNotFoundExtractionError(self.destination)
+        elif not bool(self.default_value) and self.required:
+            raise RequiredFieldNotFoundExtractionError(self.destination)
+        return self.default_value
+
```
drjova: it's better to use the guard clauses for these extractors https://www.davidhayden.me/blog/guard-clauses-in-computer-science

dags/iop/xml_extractors.py
ErnestaP: nowhere, it was since the begging when Ricardo wrote the extractor

dags/iop/xml_extractors.py
ErnestaP: I don't know, I didn't question, because it was already merged from Ricardo

dags/iop/xml_extractors.py
ErnestaP: I am not sure if I am managed to use it correctly

dags/iop/parser.py
```diff
@@ -0,0 +1,34 @@
...
+            ),
+        ]
+        super().__init__(extractors)
+
+    def _get_dois(self, article: ET.Element):
+        dois_raw = article.find("front/article-meta/article-id/[@pub-id-type='doi']")
+        if dois_raw is not None:
+            dois = dois_raw.text
+            self.logger.msg("Parsing dois for article", dois=dois)
+            return [dois]
```
```suggestion
    def _get_dois(self, article: ET.Element):
        dois_raw = article.find("front/article-meta/article-id/[@pub-id-type='doi']")
        if dois_raw is None:
            return None
        dois = dois_raw.text
        self.logger.msg("Parsing dois for article", dois=dois)
        return [dois]
```

tests/units/iop/test_iop_parser.py
```diff
@@ -0,0 +1,31 @@
...
+@fixture
+def parser():
+    return IOPParser()
+
+
+def test_doi(shared_datadir, parser):
+    content = (shared_datadir / "just_required_fields.xml").read_text()
+    article = ET.fromstring(content)
+    parsed_article = parser._publisher_specific_parsing(article)
+    assert parsed_article["dois"] == ["10.1088/1674-1137/ac66cc"]
```
drjova, line 17: Since we are going through this process, let's but in `common/jsonschema/scoap3.json` and let's have in the conftest a fixture for all tests with the `scoap3_schema` that reads the `scoap3.json` once. Then in the test we can test the subschema of a field for example:

```python
subschema = schema['properties']['dois']
assert validate(parsed_article['dois'], subschema) is None
```

dags/common/parsing/xml_extractors.py
```diff
@@ -76,22 +97,16 @@ def __init__(
...
-        return self.constant
+        if self.required:
+            if not check_value(self.default_value):
+                raise RequiredFieldNotFoundExtractionError(self.destination)
+            return self.default_value
+        return self.default_value
 
 
-class RequiredFieldNotFoundExtractionError(RuntimeError):
+class RequiredFieldNotFoundExtractionError(ValueError):
```
drjova: I would just extend from `Exception` you don't need to inherit any other type.

```python
class RequiredFieldNotFoundExtractionError(Exception):
    def __init__(self, missing_field):            
        super().__init__("Required filed is missing")
        self.missing_field = missing_field
```

dags/common/parsing/xml_extractors.py
```diff
@@ -76,22 +85,17 @@ def __init__(
...
-        super().__init__(destination)
-        self.constant = constant
 
-    def extract(self, _: ET.Element):
-        return self.constant
+        if self.required:
+            if not check_value(self.default_value):
+                raise RequiredFieldNotFoundExtractionError(self.destination)
+            return self.default_value
+        return self.default_value
```
drjova: @ErnestaP could you please check again the logic here?

drjova: Also add a test for this case?

dags/common/parsing/xml_extractors.py
```diff
@@ -23,18 +24,34 @@ def __init__(
...
+            return raw_value.text
+        except AttributeError:
+            return None
+
+    def _process_text_with_extra_function(self, text):
+        if text:
+            try:
+                return self.extra_function(text)
+            except Exception as e:
+                self.logger(e)
```
drjova: self.logger.error

drjova: you don’t need to pass e, please check the docs

dags/common/parsing/xml_extractors.py
```diff
@@ -54,13 +71,30 @@ def __init__(
...
+            return raw_value.get(self.attribute)
+        except AttributeError:
+            return None
+
+    def _process_attribute_with_extra_function(self, attribute):
+        if attribute:
+            try:
+                return self.extra_function(attribute)
+            except Exception:
+                self.logger.error("Error in extra function with value", text=attribute)
```
drjova:
```suggestion
                self.logger.error("Error in extra function with value", attribute=attribute)
```

drjova (approved):

PR #66, opened 2022-12-07
workflows: IOP journal doctype parsing

ErnestaP:
* Parsed journal doctype
* Created tests with  edge cases
* Running just specific field parser per test
* Deleted unnecessary class attributes from S3BucketResultObj

### DIFF
--- a/dags/iop/parser.py
+++ b/dags/iop/parser.py
@@ -16,12 +16,17 @@ class IOPParser(IParser):
     }
 
     def __init__(self) -> None:
+        self.dois = None
         self.logger = get_logger().bind(class_name=type(self).__name__)
         extractors = [
             CustomExtractor(
                 destination="dois",
                 extraction_function=self._get_dois,
                 required=True,
+            ),
+            CustomExtractor(
+                destination="journal_doctype",
+                extraction_function=self._get_journal_doctype,
             ),
         ]
         super().__init__(extractors)
@@ -33,5 +38,21 @@ class IOPParser(IParser):
         dois = node.text
         if dois:
             self.logger.msg("Parsing dois for article", dois=dois)
+            self.dois = dois
             return [dois]
-        return None
+        return
+
+    def _get_journal_doctype(self, article: ET.Element):
+        node = article.find(".")
+        value = node.get("article-type")
+        if not value:
+            self.logger.error("Article-type is not found in XML", doi=self.dois)
+            return None
+        try:
+            return self.article_type_mapping[value]
+        except KeyError:
+            self.logger.error(
+                "Unmapped article type", doi=self.dois, article_type=value
+            )
+        except Exception:
+            self.logger.error("Unknown error", doi=self.dois)
--- a/tests/units/iop/data/without_journal_doc_type.xml
+++ b/tests/units/iop/data/without_journal_doc_type.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Publishing DTD v1.1 20151215//EN" "JATS-journalpublishing1.dtd">
-<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" type="research-article" dtd-version="1.1" xml:lang="en">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" dtd-version="1.1" xml:lang="en">
     <front>
         <journal-meta>
             <journal-id journal-id-type="publisher-id">cpc</journal-id>
@@ -11,7 +11,7 @@
         </journal-meta>
         <article-meta>
             <article-id pub-id-type="publisher-id">cpc_46_8_085001</article-id>
-            <article-id pub-id-type="doi"></article-id>
+            <article-id pub-id-type="doi">10.1088/1674-1137/ac66cc</article-id>
             <article-id pub-id-type="manuscript">ac66cc</article-id>
             <title-group>
                 <article-title>
--- a/tests/units/iop/data/without_journal_doc_type_attribute_value.xml
+++ b/tests/units/iop/data/without_journal_doc_type_attribute_value.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Publishing DTD v1.1 20151215//EN" "JATS-journalpublishing1.dtd">
-<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" type="research-article" dtd-version="1.1" xml:lang="en">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" article-type="" dtd-version="1.1" xml:lang="en">
     <front>
         <journal-meta>
             <journal-id journal-id-type="publisher-id">cpc</journal-id>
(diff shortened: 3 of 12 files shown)


tests/units/iop/data/just_fields_no_text_data.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Publishing DTD v1.1 20151215//EN" "JATS-journalpublishing1.dtd">
-<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" article-type="research-article" dtd-version="1.1" xml:lang="en">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" article-type="" dtd-version="1.1" xml:lang="en">
```
drjova: why we need to change the test data?

dags/iop/parser.py
```diff
@@ -33,5 +39,22 @@ def _get_dois(self, article: ET.Element):
...
+        except KeyError:
+            self.logger.msg(
+                "Unmapped article type",
+                doi=self._doi,
+                article_type=value,
+                article_type_mapping=self.article_type_mapping,
+            )
+            return None
+        except Exception:
+            self.logger.msg("Unknown error", doi=self._doi)
```
drjova: why message?

dags/iop/parser.py
```diff
@@ -33,5 +39,22 @@ def _get_dois(self, article: ET.Element):
...
-        return None
+        return
+
+    def _get_journal_doctype(self, article: ET.Element):
+        node = article.find(".")
+        try:
+            value = node.get("article-type")
+            return self.article_type_mapping[value]
+        except KeyError:
+            self.logger.msg(
```
drjova: why message?

dags/iop/parser.py
```diff
@@ -33,5 +39,22 @@ def _get_dois(self, article: ET.Element):
         dois = node.text
         if dois:
             self.logger.msg("Parsing dois for article", dois=dois)
+            self._doi = dois
```
```suggestion
            self.dois = dois
```

tests/units/iop/test_iop_parser.py
```diff
@@ -13,19 +14,78 @@ def parser():
...
     content = (shared_datadir / "just_fields_no_text_data.xml").read_text()
     article = ET.fromstring(content)
     with raises(RequiredFieldNotFoundExtractionError):
+        parser.extractors[0].extract(article)
+
+
+def test_journal_doctype(shared_datadir, parser):
+    content = (shared_datadir / "just_required_fields.xml").read_text()
+    article = ET.fromstring(content)
+    parsed_article = parser.extractors[1].extract(article)
```
drjova: why `extractors[1]`? What if we change the order in the extractor? We will need to update all these tests :)

tests/units/iop/test_iop_parser.py
```diff
@@ -13,19 +14,78 @@ def parser():
...
+    parsed_doi = parser.extractors[0].extract(article)
+    assert parsed_doi == ["10.1088/1674-1137/ac66cc"]
 
 
 def test_no_doi(shared_datadir, parser):
     content = (shared_datadir / "no_data.xml").read_text()
     article = ET.fromstring(content)
     with raises(RequiredFieldNotFoundExtractionError):
-        parser._publisher_specific_parsing(article)
+        parser.extractors[0].extract(article)
```
drjova: ditto

dags/iop/parser.py
```diff
@@ -16,13 +16,19 @@ class IOPParser(IParser):
...
         extractors = [
             CustomExtractor(
                 destination="dois",
                 extraction_function=self._get_dois,
                 required=True,
             ),
+            CustomExtractor(
+                destination="journal_doctype",
+                extraction_function=self._get_journal_doctype,
+                default_value="",
```
drjova: Do we need a default value? `None` should be enough to remove it

dags/iop/parser.py
```diff
@@ -33,5 +38,22 @@ def _get_dois(self, article: ET.Element):
...
+            value = node.get("article-type")
+            return self.article_type_mapping[value]
+        except KeyError:
+            self.logger.error(
+                "Unmapped article type",
+                doi=self.dois,
+                article_type=value,
+                article_type_mapping=self.article_type_mapping,
+            )
+            return None
```
MJedr: same as above, no need for `None`

dags/iop/parser.py
```diff
@@ -33,5 +38,22 @@ def _get_dois(self, article: ET.Element):
...
             self.logger.msg("Parsing dois for article", dois=dois)
+            self.dois = dois
             return [dois]
-        return None
+        return
+
+    def _get_journal_doctype(self, article: ET.Element):
+        node = article.find(".")
+        try:
+            value = node.get("article-type")
```
MJedr: Is that possible that `"article-type"` won't be in XML? Maybe it would be worth adding another error message like `Article type not found in a XML`, it might be helpful later when filtering logging messages

dags/iop/parser.py
```diff
@@ -33,5 +38,22 @@ def _get_dois(self, article: ET.Element):
...
+            return self.article_type_mapping[value]
+        except KeyError:
+            self.logger.error(
+                "Unmapped article type",
+                doi=self.dois,
+                article_type=value,
+                article_type_mapping=self.article_type_mapping,
```
MJedr: I would exclude it from logging message, I think it's not going to be useful but it will decrease message readability

tests/units/iop/data/without_journal_doc_type.xml
```diff
@@ -0,0 +1,104 @@
...
+            </abstract>
+            <custom-meta-group>
+                <custom-meta xlink:type="simple">
+                    <meta-name>arxivppt</meta-name>
+                    <meta-value>2108.04010</meta-value>
+                </custom-meta>
+            </custom-meta-group>
+        </article-meta>
+    </front>
+</article>
```
MJedr: add an empty line in the end of the file

tests/units/iop/data/without_journal_doc_type_attribute_value.xml
```diff
@@ -0,0 +1,104 @@
...
+            </abstract>
+            <custom-meta-group>
+                <custom-meta xlink:type="simple">
+                    <meta-name>arxivppt</meta-name>
+                    <meta-value>2108.04010</meta-value>
+                </custom-meta>
+            </custom-meta-group>
+        </article-meta>
+    </front>
+</article>
```
MJedr: same here

dags/iop/parser.py
ErnestaP: it might be that the article type won't be in `article-type` or the value of the type is not in `article_type_mapping` mapping

dags/iop/parser.py
ErnestaP: I added it here because of the type, which was found in the XML might not be in `article_type_mapping`

MJedr: yes, yes, I get it but the question is if you really want to have that in all the logs? I imagine for example the whole batch of the files that for some reason have an incorrect `article_type` and in every log message, you have this mapping (which is imo redundant, what if it will grow?) and you have to unfold the message on kibana. But it's just a detail.

dags/iop/parser.py
```diff
@@ -33,5 +38,26 @@ def _get_dois(self, article: ET.Element):
         dois = node.text
         if dois:
             self.logger.msg("Parsing dois for article", dois=dois)
+            self.dois = dois
```
MJedr, line 41: actually, why do you need to set the value here, it's for logging later? isn't `CustomExtractor` setting the `dois` field value?

dags/iop/parser.py
ErnestaP: Understand, I just wanted to give the value which is received from xml and then just show it: LOOK IS NOT IN THE DICT, HERE YOU ARE CHECK IT. but if it gives more damage than value, I will remove it

dags/iop/parser.py
ErnestaP: CustomExtractor extractor sets the value in the final json, which is not accessible until all extractors have been run

ErnestaP: And we want to use doi value in the logs, while parsing other fields, so it means before the parser ends its job.

MJedr:
LGTM 👍

tests/units/iop/data/just_fields_no_text_data.xml
ErnestaP: it was the idea to have file wth fields without value

drjova (approved):

PR #67, opened 2022-12-12
workflows IOP: related article dois parsing

ErnestaP:
* parsed related article dois
* added tests
* ref: cern-sis/issues-scoap3#98

### DIFF
--- a/dags/iop/parser.py
+++ b/dags/iop/parser.py
@@ -17,6 +17,7 @@ class IOPParser(IParser):
 
     def __init__(self) -> None:
         self.dois = None
+        self.journal_doctype = None
         self.logger = get_logger().bind(class_name=type(self).__name__)
         extractors = [
             CustomExtractor(
@@ -28,13 +29,17 @@ class IOPParser(IParser):
                 destination="journal_doctype",
                 extraction_function=self._get_journal_doctype,
             ),
+            CustomExtractor(
+                destination="related_article_doi",
+                extraction_function=self._get_related_article_doi,
+            ),
         ]
         super().__init__(extractors)
 
     def _get_dois(self, article: ET.Element):
         node = article.find("front/article-meta/article-id/[@pub-id-type='doi']")
         if node is None:
-            return None
+            return
         dois = node.text
         if dois:
             self.logger.msg("Parsing dois for article", dois=dois)
@@ -46,13 +51,27 @@ class IOPParser(IParser):
         node = article.find(".")
         value = node.get("article-type")
         if not value:
-            self.logger.error("Article-type is not found in XML", doi=self.dois)
+            self.logger.error("Article-type is not found in XML", dois=self.dois)
             return None
         try:
-            return self.article_type_mapping[value]
+            self.journal_doctype = self.article_type_mapping[value]
+            return self.journal_doctype
         except KeyError:
             self.logger.error(
-                "Unmapped article type", doi=self.dois, article_type=value
+                "Unmapped article type", dois=self.dois, article_type=value
             )
         except Exception:
-            self.logger.error("Unknown error", doi=self.dois)
+            self.logger.error("Unknown error", dois=self.dois)
+
+    def _get_related_article_doi(self, article):
+        if self.journal_doctype not in ["corrigendum", "addendum"]:
+            return
+        node = article.find("front/article-meta/related-article[@ext-link-type='doi']")
+        try:
+            related_article_doi = node.get("href")
+            if related_article_doi:
+                self.logger.info("Adding related_article_doi.")
+                return [related_article_doi]
+        except AttributeError:
+            self.logger.error("No related article dois found", dois=self.dois)
+            return
--- a/tests/units/iop/data/related_article_dois_no_path.xml
+++ b/tests/units/iop/data/related_article_dois_no_path.xml
@@ -1,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Publishing DTD v1.1 20151215//EN" "JATS-journalpublishing1.dtd">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" article-type="correction" dtd-version="1.1" xml:lang="en">
+    <front>
+        <journal-meta>
+            <journal-id journal-id-type="publisher-id">cpc</journal-id>
+            <journal-title-group>
+                <journal-title xml:lang="en">Chinese Physics C</journal-title>
+            </journal-title-group>
+            <issn pub-type="ppub">1674-1137</issn>
+        </journal-meta>
+        <article-meta>
+            <article-id pub-id-type="publisher-id">cpc_46_8_085001</article-id>
+            <article-id pub-id-type="doi">10.1088/1674-1137/ac66cc</article-id>
+            <article-id pub-id-type="manuscript">ac66cc</article-id>
+            <title-group>
+                <article-title>
+                    Measurement of muon-induced neutron yield at the China Jinping Underground Laboratory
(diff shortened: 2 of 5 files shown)


dags/iop/parser.py
```diff
@@ -49,10 +54,24 @@ def _get_journal_doctype(self, article: ET.Element):
...
         except KeyError:
             self.logger.error(
                 "Unmapped article type", doi=self.dois, article_type=value
             )
         except Exception:
             self.logger.error("Unknown error", doi=self.dois)
+
+    def _get_related_article_doi(self, article):
+        if self.journal_doctype in ["corrigendum", "addendum"]:
+            node = article.find(
```
MJedr: actually, you can return if the `self.journal_doctype not in ["corrigendum", "addendum"]` and then do other stuff. Then no need for `return` cause in Python when you don't return you `None` by default

dags/iop/parser.py
```diff
@@ -49,10 +54,24 @@ def _get_journal_doctype(self, article: ET.Element):
...
+                "front/article-meta/related-article[@ext-link-type='doi']"
+            )
+            if node is None:
+                return
+            related_article_doi = node.get("href")
+            if related_article_doi:
+                self.logger.info("Adding related_article_doi.")
+                return [related_article_doi]
```
MJedr, line 74: why you need to cast to list, can't it be just a `doi` value if it's just one?

tests/units/iop/data/related_article_dois_no_path.xml
```diff
@@ -0,0 +1,104 @@
...
+            </abstract>
+            <custom-meta-group>
+                <custom-meta xlink:type="simple">
+                    <meta-name>arxivppt</meta-name>
+                    <meta-value>2108.04010</meta-value>
+                </custom-meta>
+            </custom-meta-group>
+        </article-meta>
+    </front>
+</article>
```
MJedr: add a new line ;)

tests/units/iop/data/related_article_dois_no_value.xml
```diff
@@ -0,0 +1,105 @@
...
+            </abstract>
+            <custom-meta-group>
+                <custom-meta xlink:type="simple">
+                    <meta-name>arxivppt</meta-name>
+                    <meta-value>2108.04010</meta-value>
+                </custom-meta>
+            </custom-meta-group>
+        </article-meta>
+    </front>
+</article>
```
MJedr: same

dags/iop/parser.py
ErnestaP: Good comment. Actually just because it was already implemented like this. Ricardo was keeping up with schema where dois is an array, but looks like he wasn't aware that related_article_doi + doi values will be put both in array. Fot his fix I can put dois as strings, however, we will need to change the code in generic parsing and dois/related_article_dois extractions in hindawi, aps and springer. Maybe this fix should be on another PR?

MJedr:
Looks good

dags/iop/parser.py
```diff
@@ -28,18 +29,22 @@ def __init__(self) -> None:
...
         node = article.find("front/article-meta/article-id/[@pub-id-type='doi']")
         if node is None:
-            return None
+            return
         dois = node.text
         if dois:
             self.logger.msg("Parsing dois for article", dois=dois)
             self.dois = dois
-            return [dois]
+            return dois
```
drjova: why you changed it from a list? If you do so you have to also rename it, `doi`

dags/iop/parser.py
```diff
@@ -49,10 +54,25 @@ def _get_journal_doctype(self, article: ET.Element):
...
+                node = article.find(
+                    "front/article-meta/related-article[@ext-link-type='doi']"
+                )
+                related_article_doi = node.get("href")
+                if related_article_doi:
+                    self.logger.info("Adding related_article_doi.")
+                    return related_article_doi
+        except AttributeError:
+            self.logger.error("No related article dois found", doi=self.dois)
+            return None
```
```suggestion
    def _get_related_article_doi(self, article):
        try:
            if self.journal_doctype not in ["corrigendum", "addendum"]:
                return
            node = article.find(
                "front/article-meta/related-article[@ext-link-type='doi']"
            )
            related_article_doi = node.get("href")
            if not related_article_doi:
                return
            self.logger.info("Adding related_article_doi.", doi=self.doi)
            return related_article_doi
        except AttributeError:
            self.logger.error("No related article dois found", doi=self.dois)
            return
```

drjova: Also we shouldn't wrap everything in the try-except, what could throw an `AttributeError`?

tests/units/iop/test_iop_parser.py
```diff
@@ -105,3 +105,46 @@ def test_journal_doctype_log_error_without_value(shared_datadir, parser):
...
+        assert cap_logs == [
+            {
+                "class_name": "IOPParser",
+                "dois": "10.1088/1674-1137/ac66cc",
+                "event": "Parsing dois for article",
+                "log_level": "info",
+            },
+            {
+                "class_name": "IOPParser",
+                "doi": "10.1088/1674-1137/ac66cc",
```
drjova: the naming should be consistent, we should choose between `doi` or `dois`

dags/iop/parser.py
ErnestaP: https://github.com/cern-sis/workflows/pull/67#discussion_r1045796997

dags/iop/parser.py
ErnestaP: no, in both possible error cases could be an attribute error:
`article.find`
`node.get`

drjova: then I would split into 2 try/except, the problem with the current now is you don't know which failed.

ErnestaP: It is enough just one, because article which is Element Tree obj, always have find method

drjova: if `find` doesn't throw an `AttributeError` why to include it then? :)

ErnestaP: It is not included 😅: https://github.com/cern-sis/workflows/blob/2874f7d835d755f20707f51ebac092cd4a9eebb2/dags/iop/parser.py#L69

ErnestaP: Before I was catching the error which should bever happen: https://github.com/cern-sis/workflows/pull/67#discussion_r1045927102

But later I realized, that actually one is enough. 
Did I understand the comment correctly?

tests/units/iop/test_iop_parser.py
```diff
@@ -105,3 +105,46 @@ def test_journal_doctype_log_error_without_value(shared_datadir, parser):
...
+def test_realted_article_dois_log_error_without_value(shared_datadir, parser):
+    with capture_logs() as cap_logs:
+        parser = IOPParser()
+        content = (shared_datadir / "related_article_dois_no_path.xml").read_text()
+        article = ET.fromstring(content)
+        parser._publisher_specific_parsing(article)
+        assert cap_logs == [
+            {
+                "class_name": "IOPParser",
+                "dois": "10.1088/1674-1137/ac66cc",
```
drjova, line 140: this is still different

dags/iop/parser.py
drjova: you changed it correclty 👍

tests/units/iop/test_iop_parser.py
ErnestaP: it has to be dois, no?

dags/iop/parser.py
drjova: @ErnestaP we can simplify the code:

```python
        if self.journal_doctype not in ["corrigendum", "addendum"]:
             return
        try:
            related_article_doi = article.find("front/article-meta/related-article[@ext-link-type='doi']").attrib['href']
            return [related_article_doi]
        except KeyError:
            self.logger.error("No related article dois found", dois=self.dois)
        
```

ErnestaP: if related_article_doi is None will we still have to return it?

drjova (approved):

PR #68, opened 2022-12-14
workflows IOP: extracted arxiv eprints

ErnestaP:
* extracted arxiv eprints from xml
* added tests
* ref: cern-sis/issues-scoap3#98

### DIFF
--- a/dags/iop/parser.py
+++ b/dags/iop/parser.py
@@ -1,8 +1,9 @@
 import xml.etree.ElementTree as ET
 
-from common.constants import ARXIV_EXTRACTION_PATTERN, ARXIV_VALIDATION_PATTERN
+from common.constants import ARXIV_EXTRACTION_PATTERN
 from common.parsing.parser import IParser
 from common.parsing.xml_extractors import CustomExtractor
+from idutils import is_arxiv
 from structlog import get_logger
 
 
@@ -87,9 +88,8 @@ class IOPParser(IParser):
         )
         try:
             arxiv_value = ARXIV_EXTRACTION_PATTERN.sub("", arxivs_value.text.lower())
-            if ARXIV_VALIDATION_PATTERN.match(arxiv_value):
+            if is_arxiv(arxiv_value):
                 return [{"value": arxiv_value}]
-            else:
-                self.logger.error("arXiv value is not correct format", dois=self.dois)
+            self.logger.error("The arXiv value is not valid.", dois=self.dois)
         except AttributeError:
             self.logger.error("No arXiv eprints found", dois=self.dois)
--- a/dags/iop/parser.py
+++ b/dags/iop/parser.py
@@ -1,5 +1,6 @@
 import xml.etree.ElementTree as ET
 
+from common.constants import ARXIV_EXTRACTION_PATTERN, ARXIV_VALIDATION_PATTERN
 from common.parsing.parser import IParser
 from common.parsing.xml_extractors import CustomExtractor
 from structlog import get_logger
@@ -32,6 +33,10 @@ class IOPParser(IParser):
             CustomExtractor(
                 destination="related_article_doi",
                 extraction_function=self._get_related_article_doi,
+            ),
+            CustomExtractor(
+                destination="arxiv_eprints",
+                extraction_function=self._get_extracted_arxiv_eprint_value,
             ),
         ]
         super().__init__(extractors)
@@ -75,3 +80,16 @@ class IOPParser(IParser):
         except AttributeError:
             self.logger.error("No related article dois found", dois=self.dois)
             return
+
+    def _get_extracted_arxiv_eprint_value(self, article):
+        arxivs_value = article.find(
+            "front/article-meta/custom-meta-group/custom-meta/meta-value"
+        )
+        try:
+            arxiv_value = ARXIV_EXTRACTION_PATTERN.sub("", arxivs_value.text.lower())
+            if ARXIV_VALIDATION_PATTERN.match(arxiv_value):
+                return [{"value": arxiv_value}]
+            else:
+                self.logger.error("arXiv value is not correct format", dois=self.dois)
+        except AttributeError:
+            self.logger.error("No arXiv eprints found", dois=self.dois)
--- a/tests/units/iop/data/arxiv_eprints.xml
+++ b/tests/units/iop/data/arxiv_eprints.xml
@@ -1,0 +1,104 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE article PUBLIC "-//NLM//DTD JATS (Z39.96) Journal Publishing DTD v1.1 20151215//EN" "JATS-journalpublishing1.dtd">
+<article xmlns:mml="http://www.w3.org/1998/Math/MathML" xmlns:xlink="http://www.w3.org/1999/xlink" article-type="research-article" dtd-version="1.1" xml:lang="en">
+    <front>
+        <journal-meta>
+            <journal-id journal-id-type="publisher-id">cpc</journal-id>
+            <journal-title-group>
+                <journal-title xml:lang="en">Chinese Physics C</journal-title>
+            </journal-title-group>
+            <issn pub-type="ppub">1674-1137</issn>
+        </journal-meta>
+        <article-meta>
+            <article-id pub-id-type="publisher-id">cpc_46_8_085001</article-id>
+            <article-id pub-id-type="doi">10.1088/1674-1137/ac66cc</article-id>
+            <article-id pub-id-type="manuscript">ac66cc</article-id>
+            <title-group>
+                <article-title>
+                    Measurement of muon-induced neutron yield at the China Jinping Underground Laboratory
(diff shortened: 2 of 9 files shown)


dags/iop/parser.py
```diff
@@ -75,3 +80,13 @@ def _get_related_article_doi(self, article):
         except AttributeError:
             self.logger.error("No related article dois found", dois=self.dois)
             return
+
+    def _get_arxiv_eprints(self, article):
+        arxivs_value = article.find(
+            "front/article-meta/custom-meta-group/custom-meta/meta-value"
+        )
+        try:
+            pattern = re.compile(r"(arxiv:|v[0-9]$)", flags=re.I)
```
MJedr: I would move this pattern to the constant `ARXIV_PATTERN` or something like that cause, for sure, you will need to reuse it somewhere in the repo. Or, at least, I would make it a property of the spider with a proper name.

dags/iop/parser.py
```diff
@@ -75,3 +80,13 @@ def _get_related_article_doi(self, article):
         except AttributeError:
             self.logger.error("No related article dois found", dois=self.dois)
             return
+
+    def _get_arxiv_eprints(self, article):
+        arxivs_value = article.find(
```
MJedr, line 86: maybe name it `extracted_arxiv_eprint_value`

dags/iop/parser.py
```diff
@@ -75,3 +80,13 @@ def _get_related_article_doi(self, article):
...
+            "front/article-meta/custom-meta-group/custom-meta/meta-value"
+        )
+        try:
+            pattern = re.compile(r"(arxiv:|v[0-9]$)", flags=re.I)
+            return [{"value": pattern.sub("", arxivs_value.text.lower())}]
```
MJedr: I would prefer to create a variable with proper formatting of an eprint  (so this `pattern.sub("", arxivs_value.text.lower())`) and then return, it will be more readable

tests/units/iop/data/arxiv_eprints.xml
```diff
@@ -0,0 +1,104 @@
...
+            </abstract>
+            <custom-meta-group>
+                <custom-meta xlink:type="simple">
+                    <meta-name>arxivppt</meta-name>
+                    <meta-value>2108.04010</meta-value>
+                </custom-meta>
+            </custom-meta-group>
+        </article-meta>
+    </front>
+</article>
```
MJedr: new lines missing :)

tests/units/iop/data/no_arxiv_eprints.xml
```diff
@@ -0,0 +1,103 @@
...
+                </p>
+            </abstract>
+            <custom-meta-group>
+                <custom-meta xlink:type="simple">
+                    <meta-name>arxivppt</meta-name>
+                </custom-meta>
+            </custom-meta-group>
+        </article-meta>
+    </front>
+</article>
```
MJedr: same

tests/units/iop/data/no_arxiv_eprints_value.xml
```diff
@@ -0,0 +1,104 @@
...
+            </abstract>
+            <custom-meta-group>
+                <custom-meta xlink:type="simple">
+                    <meta-name>arxivppt</meta-name>
+                    <meta-value></meta-value>
+                </custom-meta>
+            </custom-meta-group>
+        </article-meta>
+    </front>
+</article>
```
MJedr: same

dags/iop/parser.py
MJedr: Also, you are not checking if the extracted value matches the pattern, if you l run `re.sub` on whatever value, it will always be returned, which is incorrect, unless you don't care at this point about an incorrect arxiv and you want to raise an error later during validation

tests/units/iop/test_iop_parser.py
```diff
@@ -148,3 +148,46 @@ def test_realted_article_dois_log_error_without_value(shared_datadir, parser):
                 "log_level": "error",
             },
         ]
+
```
MJedr, line 152: maybe add a test that will check the correctness of the `re.sub`

dags/iop/parser.py
```diff
@@ -75,3 +80,17 @@ def _get_related_article_doi(self, article):
...
             self.logger.error("No related article dois found", dois=self.dois)
             return
+
+    def _get_extracted_arxiv_eprint_value(self, article):
+        arxivs_value = article.find(
+            "front/article-meta/custom-meta-group/custom-meta/meta-value"
+        )
+        try:
+            arxiv_value = ARXIV_EXTRACTION_PATTERN.sub("", arxivs_value.text.lower())
+            print(arxiv_value)
```
MJedr: leftover ;)

dags/iop/parser.py
```diff
@@ -75,3 +80,17 @@ def _get_related_article_doi(self, article):
...
+            if ARXIV_VALIDATION_PATTERN.match(arxiv_value):
+                return [{"value": arxiv_value}]
+            else:
+                self.logger.error("ARXIV value is not correct format", dois=self.dois)
```
MJedr: I think the correct form is `arXiv` or lowercase it as in second logger msg)

tests/units/iop/test_iop_parser.py
```diff
@@ -148,3 +149,90 @@ def test_realted_article_dois_log_error_without_value(shared_datadir, parser):
...
+    assert "1111.111aaa" == ARXIV_EXTRACTION_PATTERN.sub("", "1111.111aaa")
+    assert "11s1.111" == ARXIV_EXTRACTION_PATTERN.sub("", "11s1.111v1")
+    assert "1111.111wtwtwvq" == ARXIV_EXTRACTION_PATTERN.sub("", "1111.111wtwtwvqv1")
+
+
+def test_arxiv_validation_pattern():
+    assert bool(ARXIV_VALIDATION_PATTERN.match("1111.11111"))
+    assert not bool(ARXIV_VALIDATION_PATTERN.match("11ss1.111aaa"))
+    assert not bool(ARXIV_VALIDATION_PATTERN.match("1111.111v1"))
+    assert not bool(ARXIV_VALIDATION_PATTERN.match("1111.111wtwtwvq"))
```
MJedr: these test cases are good to use `pytest.parametrize`

tests/units/iop/test_iop_parser.py
```diff
@@ -148,3 +149,118 @@ def test_realted_article_dois_log_error_without_value(shared_datadir, parser):
...
+                "log_level": "error",
+            },
+        ]
+
+
+@fixture
+def garxiv_values():
+    for i in ["1111.111123", "1111.111wtwtwvqv1"]:
+        yield i
+
```
MJedr: is that needed?

MJedr (approved):
just a comment but looks good!

dags/iop/parser.py
```diff
@@ -75,3 +80,16 @@ def _get_related_article_doi(self, article):
...
             self.logger.error("No related article dois found", dois=self.dois)
             return
+
+    def _get_extracted_arxiv_eprint_value(self, article):
+        arxivs_value = article.find(
+            "front/article-meta/custom-meta-group/custom-meta/meta-value"
+        )
+        try:
+            arxiv_value = ARXIV_EXTRACTION_PATTERN.sub("", arxivs_value.text.lower())
+            if ARXIV_VALIDATION_PATTERN.match(arxiv_value):
```
drjova: I would recommend using `idUtils` for matching https://idutils.readthedocs.io/en/latest/

dags/iop/parser.py
```diff
@@ -75,3 +80,16 @@ def _get_related_article_doi(self, article):
...
+                return [{"value": arxiv_value}]
+            else:
+                self.logger.error("arXiv value is not correct format", dois=self.dois)
```
drjova: No need `else` since you have a return before.

dags/iop/parser.py
```diff
@@ -75,3 +80,16 @@ def _get_related_article_doi(self, article):
...
             return
+
+    def _get_extracted_arxiv_eprint_value(self, article):
+        arxivs_value = article.find(
+            "front/article-meta/custom-meta-group/custom-meta/meta-value"
+        )
+        try:
+            arxiv_value = ARXIV_EXTRACTION_PATTERN.sub("", arxivs_value.text.lower())
+            if ARXIV_VALIDATION_PATTERN.match(arxiv_value):
+                return [{"value": arxiv_value}]
```
drjova, line 92: why not returning the values as on `dois`?

dags/iop/parser.py
```suggestion
                self.logger.error("The arXiv value is not valid.", dois=self.dois)
```

dags/iop/parser.py
ErnestaP: That's the structure that was merged from Ricardo, otherwise we need to change generic parser

ErnestaP: there are things fix to fix, even though it's already merged. Where should I do it? This branch or another?

drjova (approved):

PR #69, opened 2022-12-15
Page nr

ErnestaP:


dags/iop/parser.py
```diff
@@ -93,3 +99,9 @@ def _get_extracted_arxiv_eprint_value(self, article):
             self.logger.error("The arXiv value is not valid.", dois=self.dois)
         except AttributeError:
             self.logger.error("No arXiv eprints found", dois=self.dois)
+
+    def _parse_page_to_int(self, value: str):
+        try:
+            return [int(value)]
```
MJedr: why do you always return lists even if it's a single value of another type? Is it something required by the next processing steps?

dags/iop/parser.py
```diff
@@ -93,3 +99,9 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+        except ValueError:
+            self.logger.error("Corrupted page_nr value", dois=self.dois)
```
MJedr: maybe add the `page_nr` value to the logging message

dags/iop/parser.py
```diff
@@ -93,3 +99,9 @@ def _get_extracted_arxiv_eprint_value(self, article):
             self.logger.error("The arXiv value is not valid.", dois=self.dois)
         except AttributeError:
             self.logger.error("No arXiv eprints found", dois=self.dois)
+
+    def _parse_page_to_int(self, value: str):
+        try:
+            return [int(value)]
+        except ValueError:
```
MJedr: add also `TypeError` (for `None`)

dags/iop/parser.py
MJedr: Also maybe it would be worth moving this function to utils (or another common module you have) cause you might use it in other places (in inspire we have the same function called `maybe_int`)

ErnestaP: Page_nr type according to schema is an array.
However, I agree, that we should move to put values in arrays in generic parsing or enhancement. But for this we need additional PR

dags/common/utils.py
```diff
@@ -49,3 +49,12 @@ def check_value(value):
...
     return False
+
+
+def parse_to_int(value, logger):
+    try:
+        return [int(value)]
+    except ValueError:
+        logger.error(f'Cannot parse "{value}" to integer')
+    except TypeError:
+        logger.error(f'Cannot parse "{value}" to integer')
```
MJedr:
```suggestion
    except (ValueError, TypeError):
        logger.error('Cannot parse to integer', value=value)
```

MJedr (approved):

dags/iop/parser.py
```diff
@@ -39,6 +40,12 @@ def __init__(self) -> None:
                 destination="arxiv_eprints",
                 extraction_function=self._get_extracted_arxiv_eprint_value,
             ),
+            AttributeExtractor(
+                destination="page_nr",
+                source="front/article-meta/counts/page-count",
+                attribute="count",
+                extra_function=lambda x: parse_to_int(x, self.logger),
```
drjova: you don't need to pass around the logger :) its module could have its own logger.

dags/common/utils.py
```diff
@@ -49,3 +49,10 @@ def check_value(value):
             return all(value)
         return False
     return False
+
+
+def parse_to_int(value, logger):
+    try:
+        return [int(value)]
```
drjova: the array logic shouldn't be part of this function, since it's related to the parser.

ErnestaP: I know, but it's the same situation as with dois, that we discussed

dags/iop/parser.py
ErnestaP: should I just print it?

dags/common/utils.py
drjova: Hmm, what do you mean? It's a new function we can add it to the extraction function

ErnestaP: I understood it wrong

dags/common/utils.py
```diff
@@ -49,3 +49,10 @@ def check_value(value):
             return all(value)
         return False
     return False
+
+
+def parse_to_int(value):
+    try:
+        return int(value)
+    except (ValueError, TypeError):
+        logger.error("Cannot parse to integer", value=value)
```
drjova, line 61: This logger doesn't accept extra params, I would suggest initialise a new one with structlog and also please update this line with the new one https://github.com/cern-sis/workflows/pull/69/files#diff-caa66940a5e0989c3d3a1a153a6658ead25d65b7f12ccef0d0efb3d46c032cecR30

drjova (approved):

PR #70, opened 2022-12-15
abstract parsing

ErnestaP:


dags/common/parsing/xml_extractors.py
```diff
@@ -24,9 +25,12 @@ def __init__(
         self.extra_function = extra_function
         self.logger = get_logger().bind(class_name=type(self).__name__)
 
-    def _get_text_value(self, raw_value):
+    def _get_extracted_value(self, node):
         try:
-            return raw_value.text
+            extracted_value = "".join(
+                [node.text or ""] + [ET.tostring(el).decode("ascii") for el in node]
```
MJedr: this is to get the children's values as well? if yes, maybe a more specific variable naming would be a good option as this trick seems strange at a first glance ;)

dags/common/parsing/xml_extractors.py
```diff
@@ -24,9 +25,12 @@ def __init__(
...
+            )
+            return WHITE_SPACES.sub("", extracted_value)
         except AttributeError:
             return None
```
MJedr, line 35: None is the default return

dags/common/utils.py
```diff
@@ -59,3 +61,16 @@ def parse_to_int(value):
         return int(value)
     except (ValueError, TypeError):
         logger.error("Cannot parse to integer", value=value)
+
+
+def preserve_cdata(article: str):
+    matches = CDATA_PATTERN.finditer(article)
+    for match in matches:
+        article = CDATA_PATTERN.sub(
+            match.group(1).replace("\\", "\\\\"), article, count=1
```
MJedr: why you need to pass `count`?

dags/common/parsing/xml_extractors.py
MJedr: `None` is default return

dags/common/utils.py
ErnestaP: because I want to change just the first occurrence of the CDATA and remove the CDATA tag, but leave the value inside of it. The values are different, but the pattern matches all CDATAs. Let's say, we have:
```
 <?CDATA $ \pm $?>
<?CDATA $(3.44 \pm 1.86_{\rm stat.}\pm $?>
<?CDATA $ 0.76_{\rm syst.})\times 10^{-4}$?>
```

So, I am taking the first CDATA (which before was the second) and leaving just the value:
```
  $ \pm $?
<?CDATA $(3.44 \pm 1.86_{\rm stat.}\pm $?>
<?CDATA $ 0.76_{\rm syst.})\times 10^{-4}$?>
```

And later same, taking the first one and keeping just the value:

```
$ \pm $?
$(3.44 \pm 1.86_{\rm stat.}\pm $
<?CDATA $ 0.76_{\rm syst.})\times 10^{-4}$?>
```
...

Without count, it will take all CDATAs and change to the first value from the matches, so in the end, I would have:
from this :
```
 <?CDATA $ \pm $?>
<?CDATA $(3.44 \pm 1.86_{\rm stat.}\pm $?>
<?CDATA $ 0.76_{\rm syst.})\times 10^{-4}$?>

```
To this:
```
 $ \pm $
 $ \pm $
 $ \pm $

```
while the correct one is:
```
$ \pm $?
$(3.44 \pm 1.86_{\rm stat.}\pm $
$ 0.76_{\rm syst.})\times 10^{-4}$

```

MJedr: Ok, I see. Maybe it would be possible to write a regex with [lookahead and lookbehind](https://stackoverflow.com/questions/2973436/regex-lookahead-lookbehind-and-atomic-groups) so that you'll get access directly to the match group that interests you?

ErnestaP: I need all of them, and I am getting all of them here. Just I need to change them accordingly to their value, which I don't know in advance

ErnestaP: M.. I am not sure. I  need to exchange the value of `<?CDATA .... ?>` with the value inside of CDATA (the one `....` ), so the pattern is the same

ErnestaP: I am actually getting all groups that interest me. The problem is to sub them, that's why I am using count

MJedr (approved):

tests/units/iop/test_iop_parser.py
```diff
@@ -293,3 +300,27 @@ def test_wrong_page_nr_value_log(shared_datadir, parser):
...
...inline-graphic ns0:href="cpc_46_8_085001_M1.jpg" ns0:type="simple" /></inline-formula>3.97) muon-induced neutrons from an 820.28-day dataset at the first phase of CJPL (CJPL-I). Based on the muon-induced neutrons, we measured the corresponding muon-induced neutron yield in a liquid scintillator to be<inline-formula xmlns:ns0="http://www.w3.org/1999/xlink"><tex-math>$(3.44 \pm 1.86_{\rm stat.}\pm $</tex-math><inline-graphic ns0:href="cpc_46_8_085001_M2.jpg" ns0:type="simple" /></inline-formula><inline-formula xmlns:ns0="http://www.w3.org/1999/xlink"><tex-math>$ 0.76_{\rm syst.})\times 10^{-4}$</tex-math><inline-graphic ns0:href="cpc_46_8_085001_M2-1.jpg" ns0:type="simple" /></inline-formula>&#956;<sup>&#8722;1</sup>g<sup>&#8722;1</sup>cm<sup>2</sup>at an average muon energy of 340 GeV. We provided the first study for such neutron background at CJPL. A global fit including this measurement shows a power-law coefficient of (0.75<inline-formula xmlns:ns0="http://www.w3.org/1999/xlink"><tex-math>$ \pm $</tex-math><inline-graphic ns0:href="cpc_46_8_085001_M3.jpg" ns0:type="simple" /></inline-formula>0.02) for the dependence of the neutron yield at the liquid scintillator on muon energy.'
```
drjova: black? and why `r`? :)

ErnestaP: because otherwise, I am getting the `DeprecationWarning: invalid escape sequence \p`, and these expressions are in the mathematical latex expressions, like `$(3.44 \pm 1.86_{\rm stat.}\pm $`

drjova: could you please splitted in multiple lines?

dags/common/utils.py
```diff
@@ -59,3 +61,17 @@ def parse_to_int(value):
...
+
+
+def preserve_cdata(article: str):
+    matches = CDATA_PATTERN.finditer(article)
+    for match in matches:
+        cdata_content_with_escaped_escape_chars = match.group(1).replace("\\", "\\\\")
+        article = CDATA_PATTERN.sub(
+            cdata_content_with_escaped_escape_chars, article, count=1
+        )
+    return article
```
drjova: regex are quite expensive, are sure we want to run this in everything? Also, let's add unit tests for this specific function.

ErnestaP: If I thinking better, maybe not. CDATA usually has mathematical expressions which can be in title/s or abstract. So, maybe it would be enough to run just on them. I will add tests

drjova: Yes please just call it for the ones that we really need.

ErnestaP: We can see in the tests that just by preserving the CDATA by using regex, we can get the value inside of it: 
https://github.com/cern-sis/workflows/blob/3f3239709d827cf0ef558cf56d485a592344a08f/tests/units/iop/test_iop_parser.py#L362

If we are not doing it before parsing to ET element, we lose the CDATA:
https://github.com/cern-sis/workflows/blob/3f3239709d827cf0ef558cf56d485a592344a08f/tests/units/iop/test_iop_parser.py#L390

Even when we are parsing it back to a string:
https://github.com/cern-sis/workflows/blob/3f3239709d827cf0ef558cf56d485a592344a08f/tests/units/iop/test_iop_parser.py#L397

MJedr: You can use [this function](https://github.com/inspirehep/inspire-utils/blob/6eb48fbf93963b4a5809318cf1fbc19f58db602b/inspire_utils/helpers.py#L113) to get text (including the content of cdata)

ErnestaP: I am doing it already:
https://github.com/cern-sis/workflows/blob/704383598780d0f667475a80a87abdc63c98db44/dags/common/parsing/xml_extractors.py#L32|

Is not preserving CDATA, so I need a regex in any case

ErnestaP: yes, but this is just an example that the XML is not getting back to the same structure as it was before parsing it to ElemenTree. In the article parsing, I am not sure do I need it, because I am taking the element until the deepest level, it means I have: inline-graphic, tex-math and etc. These should be kept, as far as I understand?( @drjova )

ErnestaP: Should I still remove tags @MJedr ?

PR #71, opened 2022-12-19
Authors

ErnestaP:


dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
         except AttributeError:
             self.logger.error("No arXiv eprints found", dois=self.dois)
+
+    def _get_authors(self, article):
+        contrib_types = article.findall(
+            "front/article-meta/contrib-group/contrib[@contrib-type='author']"
+        )
+        authors = []
+        surname = ""
+        given_names = ""
```
MJedr, line 115: why you set them here?

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+
+        for contrib_type in contrib_types:
+            try:
+                surname = contrib_type.find("name/surname").text
+            except AttributeError:
+                self.logger.error("Surname is not found in XML", dois=self.dois)
+            try:
+                given_names = contrib_type.find("name/given-names").text
+            except AttributeError:
+                self.logger.error("Given_names is not found in XML", dois=self.dois)
```
MJedr: I think I would extract it to smaller functions/helper function which extracts data

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
             self.logger.error("The arXiv value is not valid.", dois=self.dois)
         except AttributeError:
             self.logger.error("No arXiv eprints found", dois=self.dois)
+
+    def _get_authors(self, article):
+        contrib_types = article.findall(
```
MJedr, line 110: not sure if that's the best naming, actually these are extracted authors

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+                self.logger.error("Given_names is not found in XML", dois=self.dois)
+
+            reffered_ids = contrib_type.findall("xref[@ref-type='aff']")
+            affiliations = [
+                self._get_affiliation_value(article, reffered_id)
+                for reffered_id in reffered_ids
+                if self._get_affiliation_value(article, reffered_id) is not None
+            ]
+
+            if "collaboration" not in given_names.lower():
```
MJedr: how it is possible that `collaborations` will be in the author `given_names` entry? 😨

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+                author = clean_dict(
+                    {
+                        "surname": surname,
+                        "given_names": given_names,
+                        "affiliations": affiliations,
+                    }
```
MJedr: Are you aware that if there is an `AttributeError` error above and it won't be the first author from extracted authors, it will get the previous extracted name/given_names?

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+                )
+                authors.append(author)
+
+        if all(authors):
```
MJedr: I think that you are doing an unnecessary step here, I would check before if the required data (name, given_name, and affiliations) was retrieved and if not, I would raise an error. Also, I think there's no difference between returning `None` (as you do now) and empty list (default value for authors)

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+        except AttributeError:
+            self.logger.error("Referred id is not found")
+        institution = self._get_institution(article, id)
+        country = self._get_country(article, id)
+        try:
+            institution_and_country = {"country": country}
+            institution_and_country.update(
+                {"institution": ", ".join([institution, country])}
+            )
+        except TypeError:
```
MJedr: why is that, what types those values can have? And also why do you need to have a country in this dict and then the same information in an institution?

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+            self.logger.error("Cannot join institution and country to one value")
+        cleaned_institution_and_country = clean_dict(institution_and_country)
```
MJedr: I don't like it tbh, as I mentioned I would prefer to check if the value was extracted and then add it to the dict instead of iterating through the whole dictionary to remove empty keys

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+        if cleaned_institution_and_country:
+            return cleaned_institution_and_country
```
MJedr: same, you can return empty dict and it's also `None`

MJedr (commented):
1. Forgot to commit tests? ;)
2. I am not sure if you did it on purpose, but when you are error-logging things you are continuing the process cause you are not raising or returning in the funciton

dags/iop/parser.py
ErnestaP: yes, it is possible. IOP is putting collaborations in the author fields, where is given_name

ErnestaP:
_> * Forgot to commit tests? ;)_

No, tests are here:  https://github.com/cern-sis/workflows/blob/be9436b94e94a7ec581404bdb3562e64a4d70b03/tests/units/iop/test_iop_parser.py#L444-L835
_> * I am not sure if you did it on purpose, but when you are error-logging things you are continuing the process cause you are not raising or returning in the funciton_

I want to error log every time is a missing field, or cannot extract it. I will go longer and longer because the files I am using don't have all fields in all cases. So more extractors is parsing - more logs. Since every time running a test, it will run all extractors in the parser, it means: +1 extractor, +1 more log.

dags/iop/parser.py
ErnestaP: 1. I would get an error of  _UnboundLocalError: local variable referenced before assignment:_ here:
https://github.com/cern-sis/workflows/blob/be9436b94e94a7ec581404bdb3562e64a4d70b03/dags/iop/parser.py#L137-L138
2. I need to have string here: https://github.com/cern-sis/workflows/blob/be9436b94e94a7ec581404bdb3562e64a4d70b03/dags/iop/parser.py#L134

dags/iop/parser.py
ErnestaP: No, because it has to come from the same contrib_type, I will write a test for it

dags/iop/parser.py
ErnestaP: the thing is that neither the name, given_name, and neither affiliations are required. Just the field 'authors' is required. In the end, outputs such are these are also valid :
```
authors = [{"name": "Name"}]
authors = [{"name": "Name", "given_name": "Given name"}]
authors = [{"name": "Name", "given_name": "Given name", "affiliations": [{"country": "Blabla", "institution": "Bla"}]}]
authors = [{"name": "Name", "given_name": "Given name", "affiliations": [{"country": "Blabla"}]}]
authors = [{"affiliations": [{"country": "Blabla", "institution": "Bla"}]}]
```

dags/iop/parser.py
ErnestaP: 1. If the country won't be in XML, it will be None. None + string (institution ) = TypeError
2. That's the schema: institution and country have to be in the dict. Also, we are using these values, if I am not mistaken, is important in country "share statistics"

MJedr:
Yes, logging is correct, what I meant is that you continue parsing the file with invalid data for example (if the affiliation id is not found for example). Probably you have a validation step later, so it might be ok, but just wanted to make sure you do it on purpose.

dags/iop/parser.py
MJedr: And is it possible to extract given_names more precisely? If not maybe write a comment because it is super weird

dags/iop/parser.py
MJedr: I don't get it then. Let's say you have `contrib_types` with two entries. You iterate through this iterable. In the first iteration, you get an author, `surname=Test`, `given_names='Author`. You go to the second iteration, `surname` is not found for some reason, but from the previous iteration the value of the `surname` variable is `Test`. So you end up putting the name of the first author in the second author, am I missing something?

dags/iop/parser.py
MJedr: Yes, I know, but you should have in the loop to clean the variable after the iteration, don't you?

dags/iop/parser.py
MJedr: Ok, but I would handle it a different way than rewriting dict. You can add entries to the dict conditionally etc., there are many options ;).

dags/iop/parser.py
```diff
@@ -100,3 +105,80 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+                )
+                authors.append(author)
+
+        if all(authors):
+            return authors
+
+    def _get_affiliation_value(self, article, reffered_id):
+        institution_and_country = {}
+        try:
+            id = reffered_id.get("rid")
```
MJedr, line 156: what is the type of the `reffered_id`?

dags/iop/parser.py
MJedr: then maybe add to the logger `institution= institution, country= country`?

dags/iop/parser.py
ErnestaP: I am cleaning the variable even before putting to the array, just to avoid the one more iteration

dags/iop/parser.py
ErnestaP: no, it just has a 'collaborations' in the given name field. Good idea, I will add a comment. It is confusing, we saw it maybe just a month ago

dags/iop/parser.py
ErnestaP: I will try to explain. Let's take the example you gave me:
contrib_types has to entries, the first entry has everything, so the dict will look like this:
``{     "surname": 'Test',
            "given_names": 'Author',
            "affiliations": affiliations,
    }``

The second one doesn’t have surname, so it won't be in the dict will be without this value:
``{     
            "given_names": 'Name of the second one',
            "affiliations": affiliations,
    }``

Is it more clear?

dags/iop/parser.py
ErnestaP: ElementTree

dags/iop/parser.py
ErnestaP: test: https://github.com/cern-sis/workflows/blob/3adb98ad19e895cb7efa61696d97cda7e3d7ad67/tests/units/iop/test_iop_parser.py#L788-L832

dags/iop/parser.py
ErnestaP: It cleaned, because even if the surname is not in the XML, it's assigned to None

ErnestaP:
> Yes, logging is correct, what I meant is that you continue parsing the file with invalid data for example (if the affiliation id is not found for example). Probably you have a validation step later, so it might be ok, but just wanted to make sure you do it on purpose.

Yes, we will have the validation step by using schema, right now we cannot validate, because not all fields are parsed yet.

MJedr (approved):
Looks good now!

drjova (approved):

PR #72, opened 2022-12-20
workflows IOP: parsed date published

ErnestaP:
* parsed date published
* parsed year of date published as separate field, because this field is required
* added tests which are covering date_published and journal_year (which is a year from date_published)
* ref: ref: cern-sis/issues-scoap3#98

### DIFF
--- a/dags/iop/parser.py
+++ b/dags/iop/parser.py
@@ -1,10 +1,11 @@
 import xml.etree.ElementTree as ET
 
-from common.constants import ARXIV_EXTRACTION_PATTERN
+from common.constants import ARXIV_EXTRACTION_PATTERN, NODE_ATTRIBUTE_NOT_FOUND_ERRORS
 from common.parsing.parser import IParser
 from common.parsing.xml_extractors import AttributeExtractor, CustomExtractor
 from common.utils import parse_to_int
 from idutils import is_arxiv
+from inspire_utils.date import PartialDate
 from structlog import get_logger
 
 
@@ -20,6 +21,7 @@ class IOPParser(IParser):
 
     def __init__(self) -> None:
         self.dois = None
+        self.year = None
         self.journal_doctype = None
         self.logger = get_logger().bind(class_name=type(self).__name__)
         extractors = [
@@ -49,6 +51,15 @@ class IOPParser(IParser):
             CustomExtractor(
                 destination="authors",
                 extraction_function=self._extract_authors,
+                required=True,
+            ),
+            CustomExtractor(
+                destination="date_published",
+                extraction_function=self._get_date_published,
+            ),
+            CustomExtractor(
+                destination="journal_year",
+                extraction_function=self._get_journal_year,
                 required=True,
             ),
         ]
@@ -105,6 +116,48 @@ class IOPParser(IParser):
             self.logger.error("The arXiv value is not valid.", dois=self.dois)
         except AttributeError:
             self.logger.error("No arXiv eprints found", dois=self.dois)
+
+    def _get_date_element(self, article):
+        date = (
+            article.find("front/article-meta/history/date[@date-type='published']")
+            or article.find("front/article-meta/history/date[@date-type='epub']")
+            or article.find("front/article-meta/history/pub-date[@pub-type='ppub']")
+            or article.find("front/article-meta/pub-date")
+        )
+        return date
+
+    def _get_date_published(self, article):
+        date_element = self._get_date_element(article)
+        return self._get_date(date_element)
+
+    def _get_date(self, date):
+        try:
+            year = int(date.find("year").text)
+            self.year = year
+        except NODE_ATTRIBUTE_NOT_FOUND_ERRORS:
+            self.logger.error("Cannot find year of date_published in XML")
+        try:
+            month = int(date.find("month").text)
+        # was discussed with Anne: if the day is not present, we return year and month
+        # if the month value is not present - we return just year
+        # if the year is not present - the parsing should crash, because according schema
+        # the publication_year is required, which is take from date_published
+        except NODE_ATTRIBUTE_NOT_FOUND_ERRORS:
+            self.logger.error("Cannot find month of date_published in XML")
+            month = None
+        try:
+            day = int(date.find("day").text)
+        except NODE_ATTRIBUTE_NOT_FOUND_ERRORS:
+            self.logger.error("Cannot find day of date_published in XML")
+            day = None
+        try:
+            date = PartialDate(year=self.year, month=month, day=day)
+            return PartialDate.dumps(date)
+        except NODE_ATTRIBUTE_NOT_FOUND_ERRORS + (ValueError,):
+            return str(self.year)
+
+    def _get_journal_year(self, article):
+        return self.year
 
     def _extract_authors(self, article):
         contrib_types = article.findall(
--- a/tests/units/iop/test_iop_parser.py
+++ b/tests/units/iop/test_iop_parser.py
@@ -125,6 +125,16 @@ def test_journal_doctype_log_error_without_value(shared_datadir, parser):
                 "event": "Country is not found in XML",
                 "log_level": "error",
             },
+            {
+                "class_name": "IOPParser",
+                "event": "Cannot find month of date_published in XML",
+                "log_level": "error",
+            },
+            {
+                "class_name": "IOPParser",
+                "event": "Cannot find day of date_published in XML",
+                "log_level": "error",
+            },
         ]
 
 
@@ -186,6 +196,16 @@ def test_realted_article_dois_log_error_without_value(shared_datadir, parser):
             {
                 "class_name": "IOPParser",
                 "event": "Country is not found in XML",
+                "log_level": "error",
+            },
+            {
+                "class_name": "IOPParser",
+                "event": "Cannot find month of date_published in XML",
+                "log_level": "error",
+            },
+            {
+                "class_name": "IOPParser",
+                "event": "Cannot find day of date_published in XML",
                 "log_level": "error",
             },
         ]
@@ -251,6 +271,16 @@ def test_no_arxiv_eprints_value_log_error_without_value(shared_datadir, parser):
                 "event": "Country is not found in XML",
                 "log_level": "error",
             },
+            {
+                "class_name": "IOPParser",
+                "event": "Cannot find month of date_published in XML",
+                "log_level": "error",
+            },
+            {
+                "class_name": "IOPParser",
+                "event": "Cannot find day of date_published in XML",
+                "log_level": "error",
+            },
         ]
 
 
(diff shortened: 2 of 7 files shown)


dags/iop/parser.py
```diff
@@ -100,3 +111,50 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+        try:
+            year = int(date.find("year").text)
+            return year
+        except (AttributeError, TypeError):
+            self.logger.error("Cannot find year of date_published in XML")
+
+    def _get_date(self, date):
+        try:
+            return datetime.date(
+                day=self._get_day(date),
```
MJedr: are you sure that you don't want to return date if there's no `day`?

dags/iop/parser.py
```diff
@@ -100,3 +111,50 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+            return month
+        except (AttributeError, TypeError):
+            self.logger.error("Cannot find month of date_published in XML")
+
+    def _get_year(self, date):
+        try:
+            year = int(date.find("year").text)
+            return year
+        except (AttributeError, TypeError):
+            self.logger.error("Cannot find year of date_published in XML")
```
MJedr: maybe you can do a generic function for date extraction to which you pass an attribute name instead?

tests/units/iop/test_iop_parser.py
```diff
@@ -280,16 +320,62 @@ def test_wrong_page_nr_value_log(shared_datadir, parser):
         article = ET.fromstring(content)
         parsed_article = parser._publisher_specific_parsing(article)
         assert "page_nr" not in parsed_article
+        print(cap_logs)
```
MJedr: leftover

dags/iop/parser.py
```diff
@@ -46,6 +47,16 @@ def __init__(self) -> None:
                 attribute="count",
                 extra_function=lambda x: [parse_to_int(x)] if parse_to_int(x) else None,
             ),
+            CustomExtractor(
+                destination="date_published",
+                extraction_function=self._get_date_published,
+                default_value=datetime.date.today().isoformat(),
```
MJedr: you're putting today's date in case of missing any date element. Let's say the day is missing, but you have a year 2021 and a month 7. Then you are putting the current date as a published date, which is totally different than the real one. Are you sure it is what you want?

dags/iop/parser.py
ErnestaP: I want to use the function **_get_year** in https://github.com/cern-sis/workflows/pull/72/files#diff-ed56f3e39a25db58f05db2a77a2915ce407ab95f5ef6635e5a4453063032e413R57 as well, that's why I separated

dags/iop/parser.py
ErnestaP: mmm... I am not sure, actually. It makes more sense to return at least a part of the date. But then, if there is just a day, or there are just a day and month, what's then? In that case, is better to return today's date, no?

dags/iop/parser.py
ErnestaP: If the date is corrupted, I am not sure what to do:
1) if there is just one of the variables? (day || month || year)
2) if there are just two of them?

dags/iop/parser.py
MJedr: No, I'll apply the same logic as on inspire: https://github.com/inspirehep/inspire-utils/blob/2ae4722ce5b65c3d168b0870c2c19e03d0144954/inspire_utils/date.py#L38
So:
* year is mandatory,
* month is mandatory if there's a day.
And for the default date, I think you should ask Anne, but IMO it's a bad idea to set today's date.

dags/iop/parser.py
MJedr: I see, maybe you can derive `publication_year` from the `date_published` since it's a year part from this date? You can convert inspire https://github.com/inspirehep/inspire-utils/blob/2ae4722ce5b65c3d168b0870c2c19e03d0144954/inspire_utils/date.py#L38) and then just get the year attribute from it?

ErnestaP: Thanks, I can use it even for a date completion, if it's not full. On the other hand, I am not sure do I need it for year extraction. Year, month, and day are in different fields, so maybe I can just set a year as a class attrobute and easily reach it? As I did with doi:
 https://github.com/cern-sis/workflows/blob/main/dags/iop/parser.py#L22

MJedr: exactly, it's what I was thinking about

dags/iop/parser.py
ErnestaP: Talked with Anne:
1) if we don't have a day -> we leave just a year and a month
2) if we don't have a month -> we leave just a year
3) if we don't have a year -> it stops. Year from date_published is is the value of  journal_year which is required

dags/common/parsing/xml_extractors.py
```diff
@@ -119,6 +119,23 @@ def extract(self, article: ET.Element):
...
 
+class ConstantExtractor(IExtractor):
+    def __init__(self, destination, constant, required) -> None:
+        super().__init__(destination)
+        self.destination = destination
+        self.constant = constant
+        self.required = required
+
+    def extract(self, _: ET.Element):
+        print(self.constant)
```
MJedr: leftover

dags/common/parsing/xml_extractors.py
```diff
@@ -119,6 +119,23 @@ def extract(self, article: ET.Element):
...
+        return self.constant
+        # if check_value(self.constant):
+        #     return self.constant
+        # elif self.required:
+        #     raise RequiredFieldNotFoundExtractionError(self.destination)
+        # return self.constant
```
MJedr: leftovers

dags/common/parsing/xml_extractors.py
```diff
@@ -119,6 +119,23 @@ def extract(self, article: ET.Element):
         return self.default_value
 
 
+class ConstantExtractor(IExtractor):
```
pamfilos: is this used somewhere? maybe we should remove?

dags/common/parsing/xml_extractors.py
```diff
@@ -119,6 +119,23 @@ def extract(self, article: ET.Element):
...
+    def __init__(self, destination, constant, required) -> None:
```
MJedr: why sometimes you add type annotations and sometimes you do not?

dags/iop/parser.py
```diff
@@ -106,6 +117,51 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+        except AttributeError:
+            return
+
+    def _get_date_published(self, article):
+        date_element = self._get_date_element(article)
+        return self._get_date(date_element)
+
+    def _get_date(self, date):
+        day = None
+        month = None
```
MJedr: why do you need to define them here? can't you assign them to `None` in `except` statement?

dags/iop/parser.py
```diff
@@ -106,6 +117,51 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+            month = int(date.find("month").text)
+        # was discussed with Anne: if the day is not present, we return year and month
+        # if the month value is not present - we return just year
+        # if the year is not present - the parsing should crash, because according schema
+        # the publication_year is required, which is take from date_published
+        except (AttributeError, TypeError):
+            self.logger.error("Cannot find month of date_published in XML")
+        try:
+            day = int(date.find("day").text)
+        except (AttributeError, TypeError):
```
MJedr: I was thinking that since you're using those errors together a lot maybe you can assign them to the variable and then use this variable instead of 2 errors? Something like:

```
NODE_ATTRIBUTE_NOT_FOUND_ERRORS = (AttributeError, TypeError)
try:
    ...
except NODE_ATTRIBUTE_NOT_FOUND_ERRORS:
     ...
```

dags/iop/parser.py
```diff
@@ -106,6 +117,51 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+            self.logger.error("Cannot find day of date_published in XML")
+        try:
+            date = PartialDate(year=self.year, month=month, day=day)
+            return PartialDate.dumps(date)
+        except (AttributeError, ValueError, TypeError):
```
MJedr: why do you need those 3 exceptions here? `ParialDate` on init raises `ValueError` and `self.year` is defined, isn't it?

dags/iop/parser.py
ErnestaP: Actually, I can

dags/common/parsing/xml_extractors.py
ErnestaP: Because I am not sticking with one way of writing code. Have to change this trait. Thank you for pointing it out :)

dags/iop/parser.py
```diff
@@ -106,6 +117,51 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+            return str(self.year)
+
+    def _get_journal_year(self, _):
+        return self.year
```
pamfilos, line 160: do the extractors run in a specific order? will `self.year` be there if `_get_journal_year` runs before `_get_date_published`

pamfilos (changes requested):

dags/iop/parser.py
ErnestaP: Exceptions: I want to identify which part of the date is missing:  day, month, or year.

No, it doesn't raise an error, because the parser functions are run after the __init__, just when they are called in publisher specific parser: [here](https://github.com/cern-sis/workflows/blob/date_published/dags/common/parsing/parser.py#L40-L41)

I set a year in __init__, because as we discussed before, I need to reuse the year value in the journal_year field

dags/iop/parser.py
ErnestaP: yes, if it is in XML. Extractor runs in the order as it is in the extractors array:
https://github.com/cern-sis/workflows/blob/date_published/dags/common/parsing/parser.py#L40-L41

dags/common/parsing/xml_extractors.py
ErnestaP: no, I used it, but then changed the logic, and forgot to remove it...

dags/common/parsing/xml_extractors.py
MJedr: I think it would be good for this project to add it whenever possible, that would save you many errors later :)

dags/iop/parser.py
```diff
@@ -106,6 +117,51 @@ def _get_extracted_arxiv_eprint_value(self, article):
...
+        except NODE_ATTRIBUTE_NOT_FOUND_ERRORS:
+            self.logger.error("Cannot find day of date_published in XML")
+            day = None
+        try:
+            date = PartialDate(year=self.year, month=month, day=day)
+            return PartialDate.dumps(date)
+        except NODE_ATTRIBUTE_NOT_FOUND_ERRORS + (ValueError,):
+            return str(self.year)
+
+    def _get_journal_year(self, _):
```
MJedr: why you need to pass `_` arg to the function, is that because the extractor requires it?

MJedr (approved):
looks good, just a question

dags/iop/parser.py
ErnestaP: yes, during the extraction the article object is passed to every extractor: 
https://github.com/cern-sis/workflows/blob/401947d22b5135a5dadbdefdb05bdfe5a7d8eb57/dags/common/parsing/parser.py#L41

However, I don't nned the article in this case, because I am taking the year from the class attribute

pamfilos (approved):

PR #73, opened 2023-01-05
Copyright

ErnestaP:


dags/iop/parser.py
```diff
@@ -183,3 +188,29 @@ def _get_country(self, article, id):
...
+
+    def _get_copyright(self, article):
+        copyright = {}
+        extracted_year = self._extract_copyright_year(article)
+        if extracted_year:
+            copyright["year"] = extracted_year
+        extracted_statement = self._extract_copyright_statement(article)
+        if extracted_statement:
+            copyright["copyright_statement"] = extracted_statement
+        return [copyright]
```
ErnestaP, line 269: According the schema, it has to be array, even though it's just one element

tests/units/iop/data/no_copyright_statement.xml
```diff
@@ -0,0 +1,103 @@
...
+            </abstract>
+            <custom-meta-group>
+                <custom-meta xlink:type="simple">
+                    <meta-name>arxivppt</meta-name>
+                    <meta-value>2108.04010</meta-value>
+                </custom-meta>
+            </custom-meta-group>
+        </article-meta>
+    </front>
+</article>
```
ErnestaP: new line

dags/common/parsing/xml_extractors.py
```diff
@@ -28,7 +28,8 @@ def _get_text_value(self, raw_value):
         try:
             return raw_value.text
         except AttributeError:
-            return None
+            self.logger.error(f"{self.destination} is not found in XML")
```
MJedr, line 31: add a doi to the log msg if possible, cause if something breaks, you will search for the doi value  in logs probably

MJedr (commented):
looks good, just a small comment

dags/common/parsing/xml_extractors.py
ErnestaP: I am not sure if can I do it. The values attribute values are set on __init__, so if I would pass the doi in extractors, it will be always None. Does my thinking make sense?

MJedr: I see, maybe add it to the log msgs in iop parser then?

ErnestaP: Good point, and we actually have it. The parsing starts from doi parsing. The message, with doi is logged in doi extractor:
https://github.com/cern-sis/workflows/blob/79be516dfd37d4707748f58ae64a19d3b1f72bb0/dags/iop/parser.py#L68

dags/iop/parser.py
```diff
@@ -183,3 +188,29 @@ def _get_country(self, article, id):
         except AttributeError:
             self.logger.error("Country is not found in XML")
             return
+
+    def _extract_copyright_year(self, article):
```
pamfilos, line 245: could write one `_extract_text(path, field_name)` and use around, if common use case

dags/iop/parser.py
```diff
@@ -183,3 +188,29 @@ def _get_country(self, article, id):
...
+        try:
```
pamfilos: also maybe don't use attributeError and check if `article.find` is `None`

pamfilos (changes requested):

dags/iop/parser.py
ErnestaP: I like the solution for catching the AttributeError more. Before, I was using if/esle clauses, however,  `article.find(path) is nor None` were really confusing to everyone who checked the PR. Attribute error looks easier to understand, don't you think?

pamfilos (approved):

PR #80, opened 2023-01-10
Title and subtitle

ErnestaP:


.github/workflows/test-and-build.yml
```diff
@@ -43,32 +43,15 @@ jobs:
...
-          username: ${{ secrets.HARBOR_USERNAME }}
-          password: ${{ secrets.HARBOR_PASSWORD }}
-
-      - name: Build docker image
-        uses: docker/build-push-action@v2
-        with:
-          registry: registry.cern.ch
-          push: true
-          tags: registry.cern.ch/cern-sis/workflows:latest,registry.cern.ch/cern-sis/workflows:${{ github.sha }}
+    uses: cern-sis/gh-workflows/.github/workflows/build-image.yml@main
```
ErnestaP: Jimil's changes

dags/iop/parser.py
```diff
@@ -51,6 +55,18 @@ def __init__(self) -> None:
...
+                destination="title",
+                required=False,
+                source="front/article-meta/title-group/article-title",
+                extra_function=lambda x: " ".join(x.split()),
+            ),
+            TextExtractor(
+                destination="subtitle",
+                required=False,
+                source="front/article-meta/subtitle",
+                extra_function=lambda x: " ".join(x.split()),
```
ErnestaP, line 111: Wanted to remove all double spaces, new lines and etc.

dags/common/parsing/xml_extractors.py
```diff
@@ -1,10 +1,68 @@
...
+            ]
+            extracted_value = "".join(
+                values_in_regular_tags + values_math_expression_and_styling_tags
+            )
+            return WHITE_SPACES.sub("", extracted_value)
+        except AttributeError:
+            self.logger.error(f"{self.destination} is not found in XML")
+            return
+
+    def _process_text_with_extra_function(self, text):
```
ErnestaP: _process_text_with_extra_function, extract ar the same in ContentExtractor and in TextExtractor. Should I move these functions to a common file and reuse it here? O better to leave, that code would be more easily readable?

dags/common/constants.py
```diff
@@ -2,3 +2,4 @@
 
 ARXIV_EXTRACTION_PATTERN = re.compile(r"(arxiv:|v[0-9]$)", flags=re.I)
 NODE_ATTRIBUTE_NOT_FOUND_ERRORS = (AttributeError, TypeError)
+WHITE_SPACES = re.compile(r"[\n\t]{0,}" + r"\s{2,}")
```
MJedr: why it can't be one string?

dags/common/parsing/xml_extractors.py
MJedr: hm maybe you can move it to `IExtractor`? It won't be a super-pure interface (do I assume correctly that I come from an interface?), but well, firstly, it's not Java. Or maybe if you don't want to break Java good practices do a class that will implement this interface and add there this method.

dags/common/parsing/xml_extractors.py
```diff
@@ -1,10 +1,68 @@
 import xml.etree.ElementTree as ET
 
+from common.constants import WHITE_SPACES
 from common.parsing.extractor import IExtractor
 from common.utils import check_value
 from structlog import get_logger
 
 
+class ContentExtractor(IExtractor):
```
MJedr: where do you use this extractor?

ErnestaP: I will use it it Abstracts, if we will keep the regex idea

dags/common/constants.py
ErnestaP: + It is "AND" operator:
https://blog.finxter.com/python-regex-and-operator-tutorial-video/

dags/common/parsing/xml_extractors.py
```diff
@@ -14,14 +15,17 @@ def __init__(
         default_value=None,
         extra_function=lambda s: s,
         prefixes=None,
+        all_content_between_tags=False,
```
MJedr: I would prefer to change the name of this attribute to something like `with_tags_conent` or `extract_tags_content`

dags/common/parsing/xml_extractors.py
ErnestaP: Removed the Content extractor, extraction function added to TextExtractr. Now, when we want to extract everything between tags, we will need to pass flag in text_extractor:
`all_content_between_tags=True`

dags/common/parsing/xml_extractors.py
```diff
@@ -31,6 +35,20 @@ def _get_text_value(self, raw_value):
...
+    def _get_content_as_text_value(self, node):
+        try:
+            values_in_regular_tags = [node.text or ""]
+            values_math_expression_and_styling_tags = [
+                ET.tostring(el).decode("ascii") for el in node
+            ]
+            extracted_value = "".join(
+                values_in_regular_tags + values_math_expression_and_styling_tags
+            )
+            return WHITE_SPACES.sub("", extracted_value)
```
MJedr: can you add a test that will cover this case?

ErnestaP: actually, we need it just on abstract branch, removing from this one

dags/common/parsing/xml_extractors.py
ErnestaP: actually, we need it just on abstract branch, removing from this one

MJedr: you didn't remove in the end ;)

ErnestaP: sorry, no it's completely removed

MJedr (approved):

dags/iop/parser.py
drjova: do we have such cases? why we need to handle that?

PR #87, opened 2023-01-20
Schedule change

ErnestaP:


dags/common/sftp_service.py
```diff
@@ -20,16 +21,19 @@ def __init__(
...
                 return BytesIO(fl.read())
 
     def __connect(self):
-        cnopts = pysftp.CnOpts()
-        cnopts.hostkeys = None
+        if self.host == "localhost":
+            cnopts = pysftp.CnOpts()
+            cnopts.hostkeys = None
+        else:
+            cnopts = pysftp.CnOpts(knownhosts="known_hosts")
```
drjova: we shouldn't have special cases for `localhost`

drjova:
@ErnestaP if you check the documentation [pysftp](https://pysftp.readthedocs.io/en/release_0.2.9/cookbook.html) it's a context manager which is exactly for this reason, to open only one connection.


Example code (just for demo purposes):

```python

class SFTPService:
    def __init__(self, host, username, password):
        self.host = host
        self.username = username
        self.password = password

    def __enter__(self):
        self.conn = pysftp.Connection(host=self.host, username=self.username, password=self.password)
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.conn.close()
        
    def get_file(self, local_file, remote_file):
        with self.conn as something:
            something.do_something
````

and use it like:

```python
with SFTPService('example.com', 'username', 'password') as sftp:
    sftp.get_file('local_file.txt', 'remote_file.txt')
```

dags/springer/repository.py
```diff
@@ -52,6 +52,7 @@ def delete_all(self):
         self.s3.objects.all().delete()
 
     def __find_all_extracted_files(self):
+        print(self.s3.objects.filter(Prefix=self.EXTRACTED_DIR).all())
```
drjova: ?

dags/common/sftp_service.py
```diff
@@ -1,33 +1,31 @@
+import os
+import traceback
 from io import BytesIO
 
 import pysftp
 
 
-class SFTPService:
+class SFTPConnection:
```
drjova: @ErnestaP why we need two classes that do for that?

dags/common/sftp_service.py
```diff
@@ -16,33 +19,34 @@ def __init__(
...
-        conn = pysftp.Connection(
-            host=self.host,
-            username=self.username,
-            password=self.password,
-            port=self.port,
-            cnopts=cnopts,
-        )
-        if not conn.isdir(self.dir):
+    def __enter__(self):
+        # Bad practice, we need to fix it:
```
ErnestaP: should we solve it or keep it?

dags/hindawi/hindawi_file_processing.py
```diff
@@ -27,7 +27,7 @@ def hindawi_validate_record(enriched_file):
     validate(enriched_file, schema)
 
 
-@dag(start_date=airflow.utils.dates.days_ago(0))
+@dag(schedule_interval=None, start_date=airflow.utils.dates.days_ago(0))
```
drjova, line 30: do we need to keep the `start_date`?

dags/common/sftp_service.py
drjova: let's fix the code it self and we will fix it in another PR since it's infra related issue

dags/springer/dag_process_file.py
```diff
@@ -36,7 +36,7 @@ def springer_validate_record(enriched_file):
     validate(enriched_file, schema)
 
 
-@dag(start_date=airflow.utils.dates.days_ago(0))
+@dag(schedule_interval=None, start_date=airflow.utils.dates.days_ago(0))
```
drjova, line 39: do we need the `start_date`?

dags/springer/dag_pull_ftp.py
```diff
@@ -19,12 +21,21 @@ def springer_pull_ftp():
     logger = get_logger().bind(class_name="springer_pull_ftp")
 
     @task()
-    def migrate_from_ftp(
-        sftp: SFTPService = SpringerSFTPService(),
-        repo: IRepository = SpringerRepository(),
-        **kwargs
-    ):
-        return pull_ftp.migrate_from_ftp(sftp, repo, logger, **kwargs)
+    def migrate_from_ftp(repo: IRepository = SpringerRepository(), **kwargs):
```
drjova: I wouldn't change the signature of the function here, it will be inconsistent with the other publishers, we anyway passing the `sftp` instance, right?

tests/units/common/test_sftp.py
```diff
@@ -1,21 +1,36 @@
...
-from pysftp import Connection
 from pytest import raises
+from structlog.testing import capture_logs
 
 
-@patch.object(Connection, attribute="__init__", return_value=None)
-@patch.object(Connection, attribute="__del__", return_value=None)
-@patch.object(Connection, attribute="chdir")
-@patch.object(Connection, attribute="isdir", return_value=True)
+@patch.object(SFTPService, attribute="isdir", return_value=True)
```
drjova: do we need the `isdir` mock anymore?

dags/hindawi/hindawi_file_processing.py
ErnestaP: yes, we need it, otherwise, we need to put the start_date on every task:
https://airflow.apache.org/docs/apache-airflow/1.10.4/_modules/airflow/models/dag.html

dags/springer/dag_process_file.py
ErnestaP: yes, we need it, otherwise, we need to put the start_date on every task:
https://airflow.apache.org/docs/apache-airflow/1.10.4/_modules/airflow/models/dag.html

dags/springer/dag_pull_ftp.py
ErnestaP: Actually, no, we don't pass it. It is put in the default args to not initialize in the func. Is incorrect,
Referring also to the [article](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) you showed me once

dags/springer/dag_process_file.py
ErnestaP:
```def add_task(self, task):
        """
        Add a task to the DAG

        :param task: the task you want to add
        :type task: task
        """
        if not self.start_date and not task.start_date:
            raise AirflowException("Task is missing the start_date parameter")
        # if the task has no start date, assign it the same as the DAG
        elif not task.start_date:
            task.start_date = self.start_date
        # otherwise, the task will start on the later of its own start date and
        # the DAG's start date
        elif self.start_date:
            task.start_date = max(task.start_date, self.start_date)```

File: dags/common/sftp_service.py [hunk 9]
```diff
@@ -41,10 +37,39 @@ def __connect(self):
...
+        self.connection = self.__connect()
+        return self
+
+    def __exit__(self, exc_type, exc_value, tb):
+        if self.connection:
+            self.connection.close()
+        if exc_type is not None:
+            self.logger.error("Error while exiting SFTPService")
+            traceback.print_exception(exc_type, exc_value, tb)
+        return True
```
drjova: why we have to override this?

ErnestaP: is just a safety check, if the connection is set, and if yes, close it

drjova: if we inherit do we need to handle that?

ErnestaP: I am not using the inheritance anymore.
If we would inherit, I believe no, we don't need to handle it

dags/common/sftp_service.py
```diff
@@ -41,10 +37,39 @@ def __connect(self):
...
+
+    def list_files(self):
+        try:
+            return self.connection.listdir(self.dir)
+        except AttributeError:
+            raise NotConnectedException
+
+    def get_file(self, file):
+        try:
+            fl = self.connection.open(os.path.join(self.dir, file), "rb")
```
drjova:
```suggestion
            file_ = self.connection.open(os.path.join(self.dir, file), "rb")
```

dags/common/sftp_service.py
```diff
@@ -41,10 +37,39 @@ def __connect(self):
...
+    def __enter__(self):
+        self.connection = self.__connect()
+        return self
+
+    def __exit__(self, exc_type, exc_value, tb):
+        if self.connection:
+            self.connection.close()
+        if exc_type is not None:
+            self.logger.error("Error while exiting SFTPService")
+            traceback.print_exception(exc_type, exc_value, tb)
```
drjova: I'm not sure if traceback is attaching to structlog, you can do something like:

```
self.logger.error("An error occurred", exc_info=e)
```

dags/common/sftp_service.py
```diff
@@ -1,6 +1,9 @@
+import os
+import traceback
```
drjova, line 2: Do we need traceback?

ErnestaP: yes we do: https://github.com/cern-sis/workflows/blob/78c87c46b4e79fd6aa6e3d57f329e27659818091/dags/common/sftp_service.py#L50

drjova (approved):

PR #90, opened 2023-01-24
workflows: ConnectionError and HTTPError catching in Request class

ErnestaP:
* APS files processing dag: caught ConnectionError error while calling arxiv.org API;
* Hindawi fetching files from API dag: caught HTTPError 504 Server Error: Gateway Time-out for url;
* ref: https://github.com/cern-sis/issues-scoap3/issues/111;

### DIFF
--- a/dags/common/request.py
+++ b/dags/common/request.py
@@ -32,7 +32,10 @@ class Request:
                     path_segments={self.path_segments}>"
 
     @backoff.on_exception(
-        backoff.expo, requests.exceptions.RequestException, max_time=60, max_tries=3
+        backoff.expo,
+        (requests.exceptions.RequestException, requests.exceptions.HTTPError),
+        max_time=60,
+        max_tries=3,
     )
     def get_response(self):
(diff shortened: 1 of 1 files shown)


dags/common/request.py
```diff
@@ -32,7 +32,10 @@ def __repr__(self):
                     path_segments={self.path_segments}>"
 
     @backoff.on_exception(
-        backoff.expo, requests.exceptions.RequestException, max_time=60, max_tries=3
+        backoff.expo,
+        (requests.exceptions.RequestException, requests.exceptions.HTTPError),
```
drjova, line 36: There is no need for this PR https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions

drjova:
```
All exceptions that Requests explicitly raises inherit from [requests.exceptions.RequestException](https://requests.readthedocs.io/en/latest/api/#requests.RequestException).
```

PR #91, opened 2023-02-07
workfows: APS and Hindawi request errors fix.

ErnestaP:
* APS files processing dag: caught ConnectionError error while calling arxiv.org API.
* Hindawi fetching files from API dag: caught HTTPError 504 Server Error: Gateway Time-out for url.
* Problem solution: max_tries, removed the max_time (in order to try until the max_tries is reached) .
* ref: https://github.com/cern-sis/issues-scoap3/issues/111

### DIFF
--- a/dags/common/enricher.py
+++ b/dags/common/enricher.py
@@ -63,7 +63,7 @@ class Enricher(object):
         return list([primary_category] + secondary_categories)
 
     @backoff.on_exception(
-        backoff.expo, requests.exceptions.RequestException, max_time=60, max_tries=3
+        backoff.expo, requests.exceptions.RequestException, max_time=60, max_tries=5
     )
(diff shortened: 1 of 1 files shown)


ErnestaP:
It is a possible solution. Hard to test it, since the timeout pattern is not clear

dags/common/enricher.py
```diff
@@ -63,7 +63,7 @@ def _get_arxiv_categories_from_response_xml(self, xml: ET.Element):
         return list([primary_category] + secondary_categories)
 
     @backoff.on_exception(
-        backoff.expo, requests.exceptions.RequestException, max_time=60, max_tries=3
+        backoff.expo, requests.exceptions.RequestException, max_tries=5
```
drjova: this is a brute force, we should not keep a connection more than 60, it's too much. Let's just increase the retries.

PR #92, opened 2023-02-08
Fix for Springer reprocessing files

ErnestaP:


tests/units/common/test_sftp.py
```diff
@@ -23,15 +22,7 @@ def initiate_sftp_service():
     raises(DirectoryNotFoundException, initiate_sftp_service)
 
 
-def test_logs():
-    with capture_logs() as cap_logs:
+def test_error_raise():
+    with raises(Exception):
```
ErnestaP, line 26: Checking if the error is raised in __exit__.
It should be because we are returning False in __exit__:
https://github.com/cern-sis/workflows/blob/79606106d5e54b6d59f141ade34d2c885cd91303/dags/common/sftp_service.py#L55

dags/springer/parser.py
```diff
@@ -115,6 +114,15 @@ def __init__(self) -> None:
...
 
+    def get_dois(self, article: ET.Element):
+        source = "./Journal/Volume/Issue/Article/ArticleInfo/ArticleDOI"
+        doi_element = article.find(source)
+        if doi_element is not None:
+            dois = doi_element.text
+            self.logger.msg("Parsing dois for article", dois=dois)
+            self.dois = dois
+            return [dois]
+
```
drjova: this function will return only one doi, no?

tests/units/common/test_sftp.py
```diff
@@ -23,15 +22,7 @@ def initiate_sftp_service():
     raises(DirectoryNotFoundException, initiate_sftp_service)
 
 
-def test_logs():
-    with capture_logs() as cap_logs:
+def test_error_raise():
```
drjova, line 25: I don't understand why we need this?

dags/springer/parser.py
ErnestaP: Yes, we talked about this already before: because it is implemented to merge to doi arrays. It was made first in Springer parsing, before moving to enhancement. According to the schema, _dois_ field is array, wanted to change it before, but you said it's better to leave it for now

tests/units/common/test_sftp.py
ErnestaP: I tiried to explain it here: https://github.com/cern-sis/workflows/pull/92/files/76766090ae4f3fcc0a150d3a1644e49cc41acb99#r1100381724

I put the new return on condition, and wanted to test is it really throws an error.
If I woudn't return False in thi __exit__ function, it wouldn't throw an error, even if it was raised
docs: https://docs.python.org/3/reference/datamodel.html#object.__exit__

dags/springer/parser.py
ErnestaP: @drjova  should we open a new PR and change in all places that it infliences? It is not the first time this place raises questions.

dags/springer/dag_pull_ftp.py
```diff
@@ -10,7 +10,7 @@
 
 @dag(
     start_date=airflow.utils.dates.days_ago(0),
-    schedule_interval="@hourly",
+    schedule_interval="30 */3 * * *",
```
drjova, line 13: why this change?

dags/springer/parser.py
drjova: No, it's ok for the moment thanks

tests/units/common/test_sftp.py
drjova: Apologies, but still I don't understand why we need this test. It catches any error extending `Exception`, meaning almost everything

dags/springer/parser.py
drjova:
```suggestion
        if doi_element is None:
            return []
        dois = doi_element.text
        self.logger.msg("Parsing dois for article", dois=dois)
        self.dois = dois
        return [dois]

```

dags/springer/dag_pull_ftp.py
ErnestaP: because in the issue was mentioned:
Check and enable if is missing the schedule for its publisher, it should be the same as in PRODUCTION

tests/units/common/test_sftp.py
ErnestaP: I wanted to ensure that it catches whichever Exception and raises it. Before adding return False it was not raising anything, so the task was not stopped, even thought the connection so SFTP was not successful.

drjova (approved):

PR #94, opened 2023-02-15
Springer file structure

ErnestaP:


tests/integration/common/test_common.py
```diff
@@ -0,0 +1,60 @@
...
+                    }
+                }
+            }
+        )
+        assert len(repo.find_all()) == 1
+
+
+def test_pull_from_sftp_and_reprocess():
+    repo = SpringerRepository()
+    repo.delete_all()
```
ErnestaP: we can add this as part of the fixture (from Harris)

PR #95, opened 2023-02-16
Change of s3 files structure

ErnestaP:


dags/springer/repository.py
```diff
@@ -16,7 +16,7 @@ def __init__(self) -> None:
 
     def get_all_raw_filenames(self):
         return [
-            f.key.split("/")[-1]
+            "/".join(f.key.split("/")[1:])
```
ErnestaP: cutting off the part which doesn't belong to the file path in SFTP. for example:
raw/root_dir/file.zip -> root_dir/file.zip

dags/common/pull_ftp.py
```diff
@@ -36,38 +38,51 @@ def migrate_from_ftp(
...
+    return _find_files_in_zip(files_not_from_excluded_folder, repo)
 
 
-def _force_pull(sftp: SFTPService, repo: IRepository, logger: PrintLogger):
+def _force_pull(sftp: SFTPService, repo: IRepository, logger: PrintLogger, **kwargs):
     logger.msg("Force Pulling from SFTP.")
     filenames = sftp.list_files()
-    return migrate_files(filenames, sftp, repo, logger)
+    excluded_folders = kwargs["params"]["excluded_dirs"]
+    files_not_from_excluded_folder = remove_excluded_folders(
```
ErnestaP: should I move somewhere else? It repeats 3 times, but I am not sure where is a better place to do this filtering. All of these passes result in migrate_files, but I believe that migrate_files function should receive the final list of file names, and start migrating from sftp to s3 :
https://github.com/cern-sis/workflows/blob/2b82a0178554fbb2a86217f5d459a8670b83c4a0/dags/common/pull_ftp.py#L54-L59

https://github.com/cern-sis/workflows/blob/2b82a0178554fbb2a86217f5d459a8670b83c4a0/dags/common/pull_ftp.py#L79-L83

https://github.com/cern-sis/workflows/blob/2b82a0178554fbb2a86217f5d459a8670b83c4a0/dags/common/pull_ftp.py#L109-L112

drjova: Why we need a function for that? can we just pass a list of the excluded folders to the download function?

ErnestaP: migrate_files is the download function, but I am not really sure, should we just pass folder there. Because before it, we have to collect files names, and then pass to migrate_files.And we have 4 different collections of files:  take just missing files, pull force all, pull force spec.files. or reprocess files
 Different collections have logs different logs with list of files, for example:
    logger.msg("Pulling specified filenames from SFTP", filenames=filenames)
    logger.msg("Pulling specified filenames from SFTP", filenames=filenames)

Or is is too much repetitive and just filter files in migrate_files function? But then we will loose detailed logs that I mentioned before

ErnestaP: The collection of files comes from sftp function list_files. So it the begging it returns all file paths, and then we filter the ones we need. This filtered list is passed to download function (called migrate_files)

ErnestaP: because we are getting all file paths from sftp, and filtering just after we got it. I moved filtering  to migrate_files func, where the files are downloaded

dags/common/sftp_service.py
```diff
@@ -55,9 +57,15 @@ def __exit__(self, exc_type, exc_value, tb):
...
         return True
 
+    def store_files_name(self, fname):
+        self.file_names.append(re.sub(self.dir + "/", "", fname))
+
     def list_files(self):
         try:
-            return self.connection.listdir(self.dir)
+            self.connection.walktree(
+                self.dir, self.store_files_name, lambda x: x, lambda x: x, recurse=True
```
drjova: why we need the `lambda`s?

dags/common/utils.py
```diff
@@ -67,3 +67,14 @@ def extract_text(article, path, field_name, dois):
     except AttributeError:
         logger.error(f"{field_name} is not found in XML", dois=dois)
         return
+
+
+def remove_excluded_folders(filenames, excluded_folders):
```
drjova: I don't think we need this function, in the walktree we can just pass the directories to ignore and that's it

dags/springer/repository.py
drjova: is the `raw` in the ftp? Could you please give an example of the FTP structure?

dags/common/sftp_service.py
ErnestaP: yes, because all 3 callbacks are required arguments

dags/common/utils.py
ErnestaP: but this it is already in SFTP, should we change the SFTP? I have in mind, should we have to add the filtering in SFTP Service

dags/common/pull_ftp.py
drjova: what you are saying is about the existing method `list_files` right?

ErnestaP: yes

dags/springer/repository.py
ErnestaP: get_all_raw_filenames is the method of repo (S3).
raw file means zip file, and it is stored in the bucket with a path (example): raw/EPJC/ftp_PUB_19-01-29_20-02-10_EPJC.zip. Here I amn cutting off the "raw/" part.

SFTP looks similar, just: data/in/EPJC/ftp_PUB_19-01-29_20-02-10_EPJC.zip

dags/common/pull_ftp.py
drjova: if you check the [documentation](https://pysftp.readthedocs.io/en/release_0.2.9/pysftp.html#pysftp.Connection.walktree) one of the callbacks can be used to ignore directories, could you please explore this solution? If we do that we just need to pass the excluded dirs to `list_files` and nothing else.

ErnestaP: we cannot customize it like this, or we need to write our own walktree method:
https://stackoverflow.com/questions/68392997/pysftp-connection-walktree-fails-if-any-directory-dont-have-read-permission. 
But I can filter in store_files_name func.. but then we need to change the sftp service class, add an additional attribute "excluded_dirs"

dags/common/sftp_service.py
```diff
@@ -55,9 +57,22 @@ def __exit__(self, exc_type, exc_value, tb):
             return False
         return True
 
-    def list_files(self):
+    def store_files_name(self, fname):
+        self.file_names.append(re.sub(self.dir + "/", "", fname))
+
+    def list_files(self, excluded_dirs=[]):
```
ErnestaP: Not None because excluded_dirs in default params is []

drjova: We have discussed this in the past, it's not a good practise to have `mutable` data structures in the signatures

dags/common/sftp_service.py
```diff
@@ -22,6 +23,7 @@ def __init__(
         self.port = port
         self.logger = get_logger().bind(class_name=type(self).__name__)
         self.dir = dir
+        self.file_names = []
```
drjova: we don't need that

dags/common/sftp_service.py
ErnestaP: Remember, just I though it has to be the same as in params

dags/common/sftp_service.py
ErnestaP: but we have to save file paths somewhere, that the call back store_files_name could reach it

dags/common/sftp_service.py
```diff
@@ -55,9 +56,22 @@ def __exit__(self, exc_type, exc_value, tb):
             return False
         return True
 
-    def list_files(self):
+    def store_files_name(self, fname):
+        self.file_names.append(re.sub(self.dir + "/", "", fname))
+
+    def list_files(self, excluded_dirs=None):
```
drjova:
```suggestion
    def list_files(self, exclude_directories=None):
```

dags/common/pull_ftp.py
ErnestaP: updated: we don't need excluded_dirs in sftp, we can just filter it list_files by passing the variable of excluded_dirs there.

dags/common/sftp_service.py
ErnestaP: changed

dags/common/sftp_service.py
```diff
@@ -55,9 +56,24 @@ def __exit__(self, exc_type, exc_value, tb):
...
+                recurse=True,
+            )
+            if excluded_dirs:
+                filtered_files = []
+                for excluded_dir in excluded_dirs:
+                    filtered_files += [
+                        file for file in file_names if excluded_dir not in file
+                    ]
+                return filtered_files
+            return file_names
```
drjova: That's good, we can simplify it a lot by doing something like the following:

```python

    def list_files(self, exclude_directories=None):
        try:
            file_names = []
            self.connection.walktree(
                self.dir,
                lambda filename: append_file_if_not_in_excluded_directory(
                    filename, exclude_directories, file_names
                ),
                lambda directory: directory,
                lambda unknown: unknown,
                recurse=True,
            )
            return file_names
        except AttributeError:
            raise NotConnectedException
```

In utils.py

```python
def append_file_if_not_in_excluded_directory(filename, exclude_directories, list_of_files):
    if not exclude_directories or not (any(re.search(exclude, filename) for exclude in exclude_directories)) :
        list_of_files.append(filename)
```

And write a test about that case too in` integration/common/test_sftp.py`


```python
def test_list_files_with_exclude_directories():
    excluded_directories = ['JHEP']
    with SFTPService(dir="upload/springer/JHEP") as sftp:
        files = sftp.list_files(excluded_directories=excluded_directories)
        expected_files = []
        assert expected_files == files
```

dags/iop/dag_pull_ftp.py
```diff
@@ -9,6 +9,7 @@
 @dag(
     start_date=airflow.utils.dates.days_ago(0),
     params={
+        "excluded_dirs": [],
```
drjova:
```suggestion
        "excluded_directories": [],
```

dags/iop/repository.py
```diff
@@ -16,10 +16,7 @@ def __init__(self) -> None:
         self.s3 = S3Service(os.getenv("IOP_BUCKET_NAME", "iop"))
 
     def get_all_raw_filenames(self):
-        return [
-            os.path.basename(f.key)
-            for f in self.s3.objects.filter(Prefix=self.ZIPED_DIR).all()
-        ]
+        return [f.key for f in self.s3.objects.filter(Prefix=self.ZIPED_DIR).all()]
```
drjova: why do we need to change that?

tests/integration/common/test_common.py
```diff
@@ -0,0 +1,90 @@
...
+                        "filenames": [
+                            "JHEP/ftp_PUB_19-01-29_20-02-10_JHEP.zip",
+                            "EPJC/ftp_PUB_19-02-06_16-01-13_EPJC_stripped.zip",
+                        ],
+                        "force_from_ftp": True,
+                    },
+                }
+            }
+        )
+        assert len(springer_empty_repo.find_all()) == 2
```
drjova: let's assert with the exact expected values, not just the length - this might cover real bugs. This comments applies to the other places as well.

dags/springer/repository.py
drjova: why we can just do `replace('raw/', '')`?

drjova (changes requested):

dags/iop/repository.py
ErnestaP: f is type s3.ObjectSummary.
This place actually needs more improvement. We want to return zip file paths, without "raw/". So we need to do the same is in [Springer](https://github.com/cern-sis/workflows/blob/c30a6dd0a12aa5a360d135e49ff67caf2760ed5e/dags/springer/repository.py#L17-L21)

dags/common/sftp_service.py
```diff
@@ -55,9 +57,21 @@ def __exit__(self, exc_type, exc_value, tb):
...
 
-    def list_files(self):
+    def list_files(self, excluded_directories=None):
         try:
-            return self.connection.listdir(self.dir)
+            file_names = []
+            self.connection.walktree(
+                self.dir,
+                lambda filename: append_file_if_not_in_excluded_directory(
+                    re.sub(self.dir + "/", "", filename),
```
drjova, line 67: you don't need this 😄 if you check the `filename` actually it's like `upload/springer/JHEP/ftp_PUB_19-01-29_20-02-10_JHEP.zip` which contains already the `self.dir`.

dags/common/sftp_service.py
```diff
@@ -55,9 +57,21 @@ def __exit__(self, exc_type, exc_value, tb):
...
+                    excluded_directories,
+                    file_names,
+                ),
+                lambda directory: directory,
+                lambda unknown: unknown,
+                recurse=True,
```
drjova: it's `True` by default I believe.

dags/common/utils.py
```diff
@@ -67,3 +68,26 @@ def extract_text(article, path, field_name, dois):
...
+
+def find_extension(file):
+    if file.endswith(".xml"):
+        return "xml"
+    elif file.endswith(".pdf"):
+        return "pdf"
+    raise UnknownFileExtension(file)
+
+
+class UnknownFileExtension(Exception):
```
drjova: I would create a new module `exceptions.py` for that.

dags/iop/repository.py
```diff
@@ -31,9 +31,10 @@ def find_all(self, filenames_to_process=None):
         for file in filenames:
             last_part = os.path.basename(file)
             filename_without_extension = last_part.split(".")[0]
-            ret_dict.setdefault(filename_without_extension, defaultdict(dict))
-            extension = "xml" if self.is_meta(last_part) else "pdf"
-            ret_dict[filename_without_extension].setdefault(extension, file)
+            if filename_without_extension not in ret_dict:
```
drjova: could we add a better naming for `ret_dict`?

drjova (changes requested):
it looks good, just few comments :)

dags/common/sftp_service.py
ErnestaP: Here I am removing  "upload/springer/" 😅

drjova (approved):
LGTM 🚀 please squash

PR #96, opened 2023-02-22
Hindawi improvements

ErnestaP:


tests/units/hindawi/cassettes/test_hindawi_get_file_with_wrong_DOI.yaml
```diff
@@ -26,17 +26,17 @@ interactions:
         Content-Type:
```
ErnestaP, line 26: It got changed because we are trying to make a request again if the error was raised. Before it was 3 times, no - 6

dags/common/utils.py
```diff
@@ -12,7 +12,9 @@ def set_harvesting_interval(repo, **kwargs):
     if (
         "params" in kwargs
         and "start_date" in kwargs["params"]
+        and kwargs["params"]["start_date"]
         and "until_date" in kwargs["params"]
+        and kwargs["params"]["until_date"]
     ):
```
MJedr: maybe you can do 
```
if ( params in kwargs and  kwargs["params"].get('start_date') and kwargs["params"].get('until_date'))
```

dags/hindawi/hindawi_params.py
```diff
@@ -9,14 +9,26 @@ def __init__(
...
+        if self.record:
+            self.verb = "getrecord"
+            identifier = f"oai:hindawi.com:{self.record}"
+            params = {
+                "verb": self.verb,
+                "set": self.set,
+                "identifier": identifier,
+                "metadataprefix": self.metadataprefix,
+            }
+            return {key: value for key, value in params.items() if value}
```
MJedr: it is for a single record, right? since a majority of the logic is repeating, maybe you can have methods to prepare a different kind of payload, depending on the presence of the `self.record`

dags/hindawi/parser.py
```diff
@@ -197,7 +201,10 @@ def _get_copyright_year(self, article):
         copyright_year = re.search(
             r"[0-9]{4}", self._get_copyright_statement(article)
         ).group(0)
-        return copyright_year
+        try:
+            return int(copyright_year)
+        except ValueError:
```
MJedr: if it can be `None` by any chance you should add `TypeError`

tests/units/hindawi/cassettes/test_hindawi_get_file_with_wrong_DOI.yaml
MJedr: it is returning 301 redirecting to 403 when not found? 😮

tests/units/hindawi/test_hindawi_api_client.py
```diff
@@ -95,3 +95,11 @@ def test_hindawi_get_file_without_extension():
...
             parameters, "not_existent_extension", doi="10.1155/2019/3902819"
         )
+
+
+@mark.vcr
+def test_hindawi_get_one_record():
+    rest_api = HindawiApiClient(base_url="https://www.hindawi.com")
+    parameters = HindawiParams(record="10.1155/2022/8009789").get_params()
+    articles_metadata = rest_api.get_articles_metadata(parameters)
+    assert articles_metadata is not None
```
MJedr, line 105: I know that you have a cassette already, but maybe you can assert for anything in `articles_metadata`, a field, a number of results etc.

ErnestaP: it's an XML response 😬 (leaving it as it is, since we talked about it - we have the response written in a cassette)

tests/units/hindawi/cassettes/test_hindawi_get_file_with_wrong_DOI.yaml
ErnestaP: Apparently 😅

dags/common/utils.py
```diff
@@ -9,11 +9,7 @@
 
 
 def set_harvesting_interval(repo, **kwargs):
-    if (
-        "params" in kwargs
-        and "start_date" in kwargs["params"]
-        and "until_date" in kwargs["params"]
-    ):
+    if ("params" in kwargs and kwargs["params"].get('start_date') and kwargs["params"].get('until_date')):
```
MJedr: Did pre-commit enable that long line? Maybe it's better to break it

MJedr (approved):
just a small comment, apart of that looks good 👍

PR #97, opened 2023-02-28
workflows: changed pysftp to paramiko

ErnestaP:
* Pysftp library is not maintained since 2016.
* Filed download with patamiko is 3 times faster comparing with pysftp.
* Changed pysftp to paramiko for sftp connection.
* ref: https://github.com/cern-sis/issues-scoap3/issues/123


drjova (approved):
@ErnestaP thank you it looks good!

PR #98, opened 2023-03-01
force pull springer files

ErnestaP:


tests/integration/springer/test_dag_pull_ftp.py
```diff
@@ -52,50 +52,134 @@ def test_dag_run(dag: DAG):
     assert repo.find_all() == expected_files
 
 
-def test_dag_migrate_from_FTP():
+def test_dag_trigger_file_processing():
+    repo = SpringerRepository()
+    assert list(map(lambda x: x["xml"], repo.find_all())) == trigger_file_processing(
```
MJedr: why this map again :<<
you can simply do `[x['xml'] for x in repo.find_all()]`

tests/integration/springer/test_dag_pull_ftp.py
```diff
@@ -52,50 +52,134 @@ def test_dag_run(dag: DAG):
...
+        "springer", repo, get_logger().bind(class_name="test_logger")
+    )
+
+
+def __test_dagrun_state(dagrun: DagRun):
```
MJedr, line 62: why you need it as a private function?

dags/common/pull_ftp.py
```diff
@@ -40,13 +40,20 @@ def migrate_from_ftp(
...
-    if "force_pull" in params and params["force_pull"]:
-        return _force_pull(sftp, repo, logger, **kwargs)
-    elif (
+    force_pull_specific_files = (
         "filenames_pull" in params
         and params["filenames_pull"]["enabled"]
         and params["filenames_pull"]["force_from_ftp"]
-    ):
+    )
+    force_pull_all_files = (
```
MJedr, line 48: I am just thinking, do you really need `force_pull` param? Is it happening only when you add `excluded_directories`?

tests/integration/springer/test_dag_pull_ftp.py
ErnestaP: because it's used inside the test:
https://github.com/cern-sis/workflows/blob/5c8b56a04083118b8a4ecd414ea5192e6d0e00bb/tests/integration/springer/test_dag_pull_ftp.py#L37

dags/common/pull_ftp.py
ErnestaP: no, let me explain:
* `force_pull` is true, will force-pull all files in the SFTP.
* `force_pull` is false, just the missing files are pulled.
* If there is at least one dir in the excluded_directories, this dir will be ignored, but files from all files which are missing from the rest of the folder will be downloaded to S3. 
* For example, we have 3 folder in sftp: folder1, folder2, and folder3.

`force_pull` is **false**. 
`exluded_dir` is ["folder1"] 
In this case, missing files just from folder2 and folder3 will be downloaded. 

`force_pull` is **false**
If the `excluded_dirs` is []
The missing files from folder1, folder2, and folder3 will be downloaded again.

`force_pull` is **true**. 
`exluded_dir` is ["folder1"] 
In this case, ALL files just from folder2 and folder3 will be downloaded again (and will rewrite the files in S3 if they were already downloaded). 

`force_pull` is **true**
If the `excluded_dirs` is []
ALL files from folder1, folder2, and folder3 will be downloaded again (and will rewrite the files in S3 if they were already downloaded).

MJedr: thanks for the explenation :)

MJedr (approved):

PR #101, opened 2023-03-09
workflows IOP file processing 

ErnestaP:
* Created DAG for files processing with: enrichment, enhancement, validation.
* Ignored DS_Strore file from sftp.
* Added tests.
* Separated copyright fields in IOP parser.

### DIFF
--- a/dags/iop/dag_pull_ftp.py
+++ b/dags/iop/dag_pull_ftp.py
@@ -19,6 +19,16 @@ def iop_pull_ftp():
 
     @task()
     def migrate_from_ftp(repo=IOPRepository(), sftp=IOPSFTPService(), **kwargs):
+        params = kwargs["params"]
+        specific_files = (
+            "filenames_pull" in params
+            and params["filenames_pull"]["enabled"]
+            and not params["filenames_pull"]["force_from_ftp"]
+        )
+        if specific_files:
+            specific_files_names = pull_ftp.reprocess_files(repo, logger, **kwargs)
+            return specific_files_names
+
         with sftp:
             return pull_ftp.migrate_from_ftp(sftp, repo, logger, **kwargs)
 
--- a/dags/iop/iop_process_file.py
+++ b/dags/iop/iop_process_file.py
@@ -1,12 +1,63 @@
+import base64
+import xml.etree.ElementTree as ET
+
 import airflow
+import requests
 from airflow.decorators import dag, task
+from common.enhancer import Enhancer
+from common.enricher import Enricher
+from iop.parser import IOPParser
+from jsonschema import validate
 
 
-@dag(start_date=airflow.utils.dates.days_ago(0))
+def iop_parse_file(**kwargs):
+    if "params" not in kwargs or "file" not in kwargs["params"]:
+        raise Exception("There was no 'file' parameter. Exiting run.")
+    encoded_xml = kwargs["params"]["file"]
+    xml_bytes = base64.b64decode(encoded_xml)
+    xml = ET.fromstring(xml_bytes.decode("utf-8"))
+
+    parser = IOPParser()
+    parsed = parser.parse(xml)
+
+    return parsed
+
+
+def iop_enhance_file(parsed_file):
+    return Enhancer()("IOP", parsed_file)
+
+
+def iop_enrich_file(enhanced_file):
+    return Enricher()(enhanced_file)
+
+
+def iop_validate_record(enriched_file):
+    schema = requests.get(enriched_file["$schema"]).json()
+    validate(enriched_file, schema)
+
+
+@dag(schedule_interval=None, start_date=airflow.utils.dates.days_ago(0))
 def iop_process_file():
     @task()
-    def start():
-        pass
+    def parse_file(**kwargs):
+        return iop_parse_file(**kwargs)
+
+    @task()
+    def enhance_file(parsed_file):
+        return iop_enhance_file(parsed_file)
+
+    @task()
+    def enrich_file(enhanced_file):
+        return iop_enrich_file(enhanced_file)
+
+    @task()
+    def validate_record(enriched_file):
+        iop_validate_record(enriched_file)
+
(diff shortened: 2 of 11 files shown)


dags/iop/dag_pull_ftp.py
```diff
@@ -19,6 +19,16 @@ def iop_pull_ftp():
 
     @task()
     def migrate_from_ftp(repo=IOPRepository(), sftp=IOPSFTPService(), **kwargs):
+        params = kwargs["params"]
+        specific_files = (
+            "filenames_pull" in params
+            and params["filenames_pull"]["enabled"]
+            and not params["filenames_pull"]["force_from_ftp"]
```
MJedr, line 26: the keys `enabled` and `force_from_ftp` are always in params?

dags/iop/iop_process_file.py
```diff
@@ -1,12 +1,63 @@
...
+    if "params" in kwargs and "file" in kwargs["params"]:
+        encoded_xml = kwargs["params"]["file"]
+        xml_bytes = base64.b64decode(encoded_xml)
+        xml = ET.fromstring(xml_bytes.decode("utf-8"))
+
+        parser = IOPParser()
+        parsed = parser.parse(xml)
+
+        return parsed
+    raise Exception("There was no 'file' parameter. Exiting run.")
```
MJedr: maybe you can revert the condition (`if "params" in kwargs and "file" in kwargs["params"]`) and raise it at the beginning of the function?

dags/iop/parser.py
```diff
@@ -83,9 +79,12 @@ def __init__(self) -> None:
                 required=True,
             ),
             CustomExtractor(
-                destination="copyright",
-                extraction_function=self._get_copyright,
-                required=True,
+                destination="copyright_year",
+                extraction_function=self._extract_copyright_year,
```
MJedr, line 83: where is this method now?

tests/integration/iop/test_dag_process_file.py
```diff
@@ -0,0 +1,149 @@
...
+@fixture
+def article(parser):
+    data_dir = "./data/iop/"
+    test_file = "2022-07-30T03_02_01_content.zip"
+
+    def extract_zip_to_article(zip_filename):
+        with ZipFile(zip_filename, "r") as zip_file:
+            return [
+                zip_file.read(filename)
+                for filename in map(lambda x: x.filename, zip_file.filelist)
```
MJedr: map? :((( it's unpythonic, see the discussion [https://stackoverflow.com/questions/1247486/list-comprehension-vs-map](there) 😝 

But jokes aside, in this example it's reasonable to use a list comprehension cause you don't need lambda to perform any kind of operation on the list. You can also split this return in 2 statements for readability.

tests/integration/iop/test_dag_process_file.py
```diff
@@ -0,0 +1,149 @@
...
+    }
+    assertListEqual(
+        {
+            "arxiv_eprints": [
+                {"value": "2112.01211", "categories": list(set(["hep-th", "hep-ph"]))}
+            ],
+            "$schema": "http://repo.qa.scoap3.org/schemas/hep.json",
+        },
+        iop_enrich_file(input_article),
+    )
```
MJedr, line 149: just a question, cause don't know if it's possible: can't you test the whole dag processing pipeline `iop_process_file`?

dags/iop/parser.py
ErnestaP: https://github.com/cern-sis/workflows/blob/5f3a04512c90045cc1898f3bbfe3830160e6ba02/dags/iop/parser.py#L310-L316

ErnestaP: I had a discussion about it with Harris. The initial idea was to have all subfields parsed separately and then joined into one field after all processing steps are done (generic parsing, enchantment, and enrichment. In this PR we have the publisher-specific parser). 
However, I switched to the idea, that maybe it's better to join these fields in the final field already here because, in the [enhancer](https://github.com/cern-sis/workflows/blob/76307b8c4085fd7a41634efb8ec105423bad2542/dags/common/enhancer.py#L15-L23), we are creating unnecessary dict keys and then deleting them if the field value is empty. 
However, then we are creating this big field in every publisher, rather than just to have in one place. So, in the end, we decided to keep subfields separately parsed.

tests/integration/iop/test_dag_process_file.py
ErnestaP: I actually cab but just to check the state of the DAG: is it "success", "running", "queued", "failed".
Something similar to this:
https://github.com/cern-sis/workflows/blob/5f3a04512c90045cc1898f3bbfe3830160e6ba02/tests/units/aps/test_trigger_files_processing.py#L59-L76

Since we are not saving the ouput of this DAG is hard to test it differently, that's why I am testing every step separately

dags/iop/dag_pull_ftp.py
ErnestaP: if we want to harvest specific files or reprocess them, then yes, it has to be always in Params, if the publisher's source is SFTP/FTP

dags/iop/parser.py
MJedr: ok, sure, but you reference the method by `self` and it's in `Enhancer`

ErnestaP: the methods referenced by self always are in the parser, because are parser-specific methods. 
I wanted to remove **copyright field** construction from Enhancer and just keep it as part of the publisher-specific parser method. But we moved from this idea. Now we are keeping subfields of copyright parsing: **copyright_year** and **copyright_statement** parsed separately and just joining them in copyright, in the Enhancer.

The Enhancer change was never done and it had to be on separate PR because it would have affected all publishers. But it won't be done at all, because we moved from the idea of having fields joined in the publisher-specific parser. 

Is it more clear?

MJedr (approved):

PR #102, opened 2023-03-14
Oup parsing doi

ErnestaP:


dags/common/utils.py
```diff
@@ -149,3 +150,16 @@ def preserve_cdata(article: str):
 
 def parse_to_ET_element(article: str):
     return ET.fromstring(preserve_cdata(article))
+
+
+def parse_without_names_spaces(xml: str):
+    it = ET.iterparse(StringIO(xml))
+    for _, el in it:
+        _, _, el.tag = el.tag.rpartition("}")
```
MJedr: maybe `el.tag = el.tag.rpartition("}")[-1]`?

dags/common/utils.py
```diff
@@ -149,3 +150,16 @@ def preserve_cdata(article: str):
...
+    root = it.root
+    return root
+
+
+def get_text_value(element: ET.Element):
+    if element is not None:
```
MJedr, line 164: `if element:` won't work?

dags/oup/parser.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def _get_dois(self, article: ET.Element):
+        source = ".front/article-meta/article-id/[@pub-id-type='doi']"
+        doi_element = article.find(source)
+        dois = get_text_value(doi_element)
+        if not dois:
+            return []
+        self.logger.msg("Parsing dois for article", dois=dois)
+        self.dois = dois
+        return [dois]
```
MJedr: aren't dois a string? if yes, shouldn't you split? or maybe it's just one doi, then shouldn't you rename var to `doi`?

dags/common/utils.py
```diff
@@ -149,3 +150,16 @@ def preserve_cdata(article: str):
 
 def parse_to_ET_element(article: str):
     return ET.fromstring(preserve_cdata(article))
+
+
+def parse_without_names_spaces(xml: str):
```
MJedr, line 155: you need it for tests only? if yes, maybe you can move it to the test/utils or something?

tests/units/oup/test_oup_parser.py
```diff
@@ -0,0 +1,50 @@
...
+@fixture()
+def parsed_articles(parser, valid_articles):
+    yield [parser._publisher_specific_parsing(article) for article in valid_articles]
+
+
+def test_dois(parsed_articles):
+    dois = ["10.1093/ptep/ptac108", "10.1093/ptep/ptac120", "10.1093/ptep/ptac113"]
+    dois_parsed_article = [article["dois"][0] for article in parsed_articles]
+    for doi, article_doi in zip(sorted(dois), sorted(dois_parsed_article)):
+        assert article_doi == doi
```
MJedr: basically you can do `set(dois) == set(dois_parsed_article)` instead

dags/common/utils.py
ErnestaP: No, because the bool(element) will return False. I know, it is really weird 😅

dags/oup/parser.py
ErnestaP: I should split doi? What do you mean?
Actually, you can get here just one doi.
Dois because t keep it consistent with all other publishers

dags/common/utils.py
ErnestaP: no, I will use it also in the file processing  DAG

dags/common/utils.py
ErnestaP: review

MJedr: it's ok, sometimes it's like that ;)

dags/oup/parser.py
MJedr: 1. You do `get_text_value(doi_element)`
2. If there are multiple dois, you get something like `"10.1103/PhysRevLett.7.394, 10.1103/PhysRevLett.8.928"` (cause you get text) so I am just wondering how it is possible that you can return an array without splitting the result of  `get_text_value(doi_element)`

ErnestaP: there is just one doi by field, it means get_text_value(doi_element) will return just one doi.
One article can have just one doi.  Array - just because to align with other deployed publishers. It has to be fixed later 😅

MJedr: ok then 👍 so maybe rename the variable to `doi` instead of `dois`, then it would be less ambiguous ;)

ErnestaP: but the field name in the schema called dois 😅

MJedr (approved):

PR #103, opened 2023-03-15
Oup parsing journal doctype

ErnestaP:


ErnestaP:
Will be ready for review, after https://github.com/cern-sis/workflows/pull/102 is merged, for less confusion

dags/oup/parser.py
```diff
@@ -34,3 +39,19 @@ def _get_dois(self, article: ET.Element):
...
+    def _get_journal_doctype(self, article: ET.Element):
+        self.logger.msg("Parsing journal doctype for article", dois=self.dois)
+        journal_doctype_raw = article.find("front/..").get("article-type")
+        if not journal_doctype_raw:
+            return
+        try:
+            journal_doctype = self.article_type_mapping[journal_doctype_raw]
+        except KeyError:
+            journal_doctype = "other"
+        if "other" in journal_doctype:
```
MJedr: does it happen anytime except of the `KeyError`? If no, maybe you can remove `if` statement

tests/units/oup/test_oup_parser.py
```diff
@@ -36,14 +35,40 @@ def test_dois(parsed_articles):
...
     article_name = "ptac108_without_doi_value.xml"
     with open(shared_datadir / article_name) as file:
-        content = ET.fromstring(file.read())
+        content = parse_without_names_spaces(file.read())
         with raises(RequiredFieldNotFoundExtractionError):
             parser._publisher_specific_parsing(content)
+
+
+def test_journal_doc_types(parsed_articles):
+    doc_types = ["article", "article", "article"]
```
MJedr: would be nice to have an article in other type there

dags/oup/parser.py
ErnestaP: Yes, just when the type is not found in article type mapping

tests/units/oup/test_oup_parser.py
ErnestaP: I can try to find one

ErnestaP: managed to find just 1

dags/oup/parser.py
ErnestaP: Can it be a different exception in this case?

MJedr: no, I would just move logger warning to the exception handling and remove `if` clause

MJedr (approved):

PR #104, opened 2023-03-15
Oup parsing arxiv

ErnestaP:


dags/oup/parser.py
```diff
@@ -54,8 +58,20 @@ def _get_journal_doctype(self, article: ET.Element):
...
             )
         return journal_doctype
+
+    def _get_arxiv_eprints(self, article: ET.Element):
+        arxivs_raw = get_text_value(
+            article.find("front/article-meta/article-id/[@pub-id-type='arxiv']")
+        )
+        if not arxivs_raw:
+            return
+        arxiv_eprint = arxivs_raw.lower().replace("arxiv:", "")
```
MJedr, line 74: are you sure arxiv id is always like that? on inspire we have some different versions https://github.com/inspirehep/inspire-schemas/blob/32d9168cb17d6388d67aad311dff93dac99b15b0/inspire_schemas/utils.py#L102-L105
and recently we've been relaxing our schema to accept another pattern

tests/units/oup/test_oup_parser.py
```diff
@@ -109,3 +109,23 @@ def other_doc_type_article(shared_datadir, parser):
...
+
+def test_arxiv(parsed_articles):
+    doc_types = [
+        {"value": "2111.09468"},
+        {"value": "2204.01249"},
+        {"value": "2207.02498"},
+        {"value": "2205.14599"},
+    ]
+    for doc_type, article_doc_type in zip(doc_types, parsed_articles):
+        print(article_doc_type["arxiv_eprints"])
```
MJedr: print :)

dags/oup/parser.py
ErnestaP: The format of arxiv is not specified in this change. It is just "arxiv:" part removal. If this part is not in arxiv string, the string of arxiv value won't be changed

MJedr (approved):

MJedr (approved):

PR #109, opened 2023-03-20
Workflows: OUP authors parsing

ErnestaP:
* Authors parsing.
* Added tests.

### DIFF
--- a/dags/oup/parser.py
+++ b/dags/oup/parser.py
@@ -36,6 +36,11 @@ class OUPParser(IParser):
                 destination="arxiv_eprints",
                 extraction_function=self._get_arxiv_eprints,
             ),
+            CustomExtractor(
+                destination="authors",
+                extraction_function=self._get_authors,
+                required=True,
+            ),
         ]
         super().__init__(extractors)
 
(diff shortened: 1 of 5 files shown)


dags/oup/parser.py
```diff
@@ -53,3 +58,42 @@ def _get_journal_doctype(self, article: ET.Element):
...
+                    ),
+                    "country": get_text_value(
+                        affiliation.find(
+                            "country",
+                        )
+                    ).capitalize(),
+                }
+                for affiliation in affiliations
+            ]
+            if not surname and not given_names and not email and not full_affiliation:
```
MJedr:
```
if not all([surname, given_names, email, full_affiliation]):
```

dags/oup/parser.py
```diff
@@ -53,3 +58,42 @@ def _get_journal_doctype(self, article: ET.Element):
...
+                    "institution": get_text_value(
+                        affiliation.find(
+                            "institution",
+                        )
+                    ),
+                    "country": get_text_value(
+                        affiliation.find(
+                            "country",
+                        )
+                    ).capitalize(),
```
MJedr: if `get_text_value` returns `None` you'll get an error

dags/oup/parser.py
MJedr: actually maybe move it before building affiliation

ErnestaP: `if not all([surname, given_names, email, full_affiliation]):` won't work, because is an array of dicts, which will give an error ` TypeError: 'builtin_function_or_method' object is not subscriptable`

MJedr: well, it should work with dicts:
```
>>> a = {'test': 2}
>>> b = {'another': 3}
>>> c = {'yet': 5}
>>> all([a, b, c])
True
```
you need two brackets, it occurs when you add just a square bracket
```
>>> all[a, b, c, 7]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'builtin_function_or_method' object is not subscriptable
```

MJedr (approved):

PR #110, opened 2023-03-20
Workflows OUP: published date parsing

ErnestaP:
* Parsed date published.
* Added test.

### DIFF
--- a/dags/oup/parser.py
+++ b/dags/oup/parser.py
@@ -57,7 +57,7 @@ class OUPParser(IParser):
                 source="front/article-meta/title-group/article-title",
                 all_content_between_tags=True,
             ),
-            TextExtractor(
+            CustomExtractor(
                 destination="date_published",
                 extraction_function=self._get_published_date,
             ),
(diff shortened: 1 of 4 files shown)


dags/oup/parser.py
```diff
@@ -136,3 +141,21 @@ def _get_authors(self, article: ET.Element):
...
                 )
         return authors
+
+    def __get_date(self, article):
+        year = article.find("year").text
+        month = article.find("month").text
+        day = article.find("day").text
+        if year and month and day:
+            return datetime.date(
+                day=int(day), month=int(month), year=int(year)
```
MJedr, line 151: you don't need to cast to int

dags/oup/parser.py
```diff
@@ -136,3 +141,21 @@ def _get_authors(self, article: ET.Element):
                     }
                 )
         return authors
+
+    def __get_date(self, article):
```
MJedr: I won't make it private because previously you were not using private methods, so let's be consistent

dags/oup/parser.py
ErnestaP: I need it :
https://docs.python.org/3/library/datetime.html#date-objects

MJedr (approved):

PR #111, opened 2023-03-21
Workflows OUP: journal title parsing

ErnestaP:


dags/oup/parser.py
```diff
@@ -56,6 +56,11 @@ def __init__(self) -> None:
                 source="front/article-meta/title-group/article-title",
                 all_content_between_tags=True,
             ),
+            CustomExtractor(
+                destination="journal_title",
+                extraction_function=self._get_journal_title,
+                required=True,
```
MJedr, line 97: required is also default prop, right?

dags/oup/parser.py
```diff
@@ -136,3 +141,27 @@ def _get_authors(self, article: ET.Element):
...
+        if journal_publisher:
+            return journal_publisher
+
+        journal_publisher = get_text_value(
+            article.find(
+                "front/journal-meta/journal-title-group/abbrev-journal-title/[@abbrev-type='publisher']"
+            )
+        )
+        if journal_publisher:
+            return journal_publisher
```
MJedr: you can do `journal_publisher  = article.find(<first path>) or article.find(<second path>)` -> if the first condition is met second is not checked and then just return `journal_publisher` since if you don't return anything you return `None`

tests/units/oup/test_oup_parser.py
```diff
@@ -313,7 +313,7 @@ def test_no_abstract_value(shared_datadir, parser):
 
 def test_titles(parsed_articles):
     titles = [
-        "Detectable electric current induced by the dark matter axion in a conductor",
+        "Detectable electric current induced by the dark matter axion in aconductor",
```
MJedr: why is that? 😄

dags/oup/parser.py
ErnestaP: we decided to remove the default prop completely because there is no sense to have default prop if the required field is missing

tests/units/oup/test_oup_parser.py
ErnestaP: :DDDDD I think it was by mistake, I deleted the space in the XML and then "mm, tests are failing  "

dags/oup/parser.py
ErnestaP: It's actually not publisher, it's journal_title_short_form, I will change the name of the var

tests/units/oup/test_oup_parser.py
ErnestaP: my bad :D I deleted the space in file, most likely

dags/oup/parser.py
ErnestaP: If you are talking about "required" in custom extractor it is False by default

tests/units/oup/test_oup_parser.py
ErnestaP: mmm actually is after the file formatting

ErnestaP: fixed: actually, I needed to change the cleaning string value, after parsing (in xml_extractors text_extractor)

tests/units/oup/test_oup_parser.py
```diff
@@ -512,6 +549,30 @@ def test_licenses_no_license(shared_datadir, parser):
...
 
+def test_no_journal_title_value(shared_datadir, parser):
+    article_name = "ptab170_no_journal_title_value.xml"
+    with open(shared_datadir / article_name) as file:
+        content = parse_without_names_spaces(file.read())
+        article = parser._publisher_specific_parsing(content)
+        article["journal_title"] == "Prog. Thezor. Exp. Phys"
+
+
+def test_no_journal_title_and(shared_datadir, parser):
```
MJedr: isn't it missing part of the test name?

MJedr (approved):

PR #112, opened 2023-03-21
Oup parsing journal issue

ErnestaP:


dags/oup/parser.py
```diff
@@ -56,6 +56,12 @@ def __init__(self) -> None:
                 source="front/article-meta/title-group/article-title",
                 all_content_between_tags=True,
             ),
+            TextExtractor(
+                destination="journal_issue",
+                source="front/article-meta/issue",
+                extra_function=lambda x: x,
```
MJedr: it's default, right?

ErnestaP: yes! 😬😄

MJedr (approved):

PR #113, opened 2023-03-22
Workflows: OUP journal volume and year parsing

ErnestaP:
* Journal volume parsing.
* Added test.

### DIFF
--- a/dags/oup/parser.py
+++ b/dags/oup/parser.py
@@ -22,6 +22,7 @@ class OUPParser(IParser):
             "addendum": "addendum",
             "editorial": "editorial",
         }
+        self.year = None
 
         extractors = [
             CustomExtractor(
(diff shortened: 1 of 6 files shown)


dags/oup/parser.py
```diff
@@ -56,6 +57,15 @@ def __init__(self) -> None:
...
                 all_content_between_tags=True,
             ),
+            CustomExtractor(
+                destination="journal_volume",
+                extraction_function=self._get_journal_volume,
+            ),
+            CustomExtractor(
+                destination="journal_year",
+                extraction_function=self._get_journal_year,
+                required=True,
```
MJedr, line 77: default is True ;)

ErnestaP: In custom extractor default is False

ErnestaP: https://github.com/cern-sis/workflows/blob/2644d9c5b223f0329b1a109c51efb5b233752bac/dags/common/parsing/xml_extractors.py#L128

MJedr (approved):

PR #116, opened 2023-03-22
Workflows OUP: parsing copyright year

ErnestaP:
* Parsing copyright year.
* Added tests.
* ref: cern-sis/issues-scoap3#96

### DIFF
--- a/dags/oup/parser.py
+++ b/dags/oup/parser.py
@@ -81,6 +81,12 @@ class OUPParser(IParser):
                 source="front/article-meta/elocation-id",
                 required=False,
             ),
+            TextExtractor(
+                destination="copyright_year",
+                source="front/article-meta/permissions/copyright-year",
+                extra_function=lambda x: int(x),
+                required=False,
+            ),
         ]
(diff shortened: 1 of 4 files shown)


dags/oup/parser.py
```diff
@@ -41,6 +45,12 @@ def __init__(self) -> None:
                 extraction_function=self._get_authors,
                 required=True,
             ),
+            TextExtractor(
+                destination="copyright_year",
+                source="front/article-meta/permissions/copyright-year",
+                extra_function=lambda x: int(x),
```
MJedr, line 87: what happens if for some reason a value can't be converted to int?

ErnestaP: error and return None. But it should not stop the parsing. 
I am handling the situations like this in the extractor:
https://github.com/cern-sis/workflows/blob/main/dags/common/parsing/xml_extractors.py#L53-L58

You have a good eye! Thanks for checking PRs in details :)

MJedr (approved):

PR #123, opened 2023-04-24
workflows: OUP ftp server connection

ErnestaP:
* Created FTP connection class, for FTP server connections.
* Set up FTP in Docker.
* Still working on tests.
* ref: https://github.com/cern-sis/issues-scoap3/issues/157

### DIFF
--- a/dags/common/ftp_service.py
+++ b/dags/common/ftp_service.py
@@ -1,0 +1,79 @@
+import ftplib
+import io
+import os
+import re
+import traceback
+
+from common.exceptions import DirectoryNotFoundException, NotConnectedException
+from common.utils import append_file_if_not_in_excluded_directory, walk_ftp
+from structlog import get_logger
+
+
+class FTPService:
+    def __init__(
+        self,
+        host="localhost",
+        username="airflow",
+        password="airflow",
+        port=2222,
+        dir="/upload",
+    ):
+        self.connection = None
+        self.host = host
+        self.username = username
+        self.password = password
+        self.port = port
+        self.logger = get_logger().bind(class_name=type(self).__name__)
+        self.dir = dir
+
+    def __connect(self):
+        ftp = ftplib.FTP(self.host)
+        ftp.login(self.username, self.password)
+        try:
+            ftp.dir(self.dir)
+        except FileNotFoundError:
+            raise DirectoryNotFoundException(
+                "Remote directory doesn't exist. Abort connection."
+            )
+        return ftp
+
+    def __enter__(self):
+        self.connection = self.__connect()
+        return self
+
+    def __exit__(self, exc_type, exc_value, tb):
+        if self.connection:
+            self.connection.close()
+        if exc_type is not None:
+            formed_exception = traceback.format_exception_only(exc_type, exc_value)
+            self.logger.error(
+                "An error occurred while exiting FTPService",
+                execption=formed_exception,
+            )
+            return False
+        return True
+
+    def list_files(self, excluded_directories=None):
+        try:
+            file_names = []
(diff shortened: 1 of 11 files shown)


dags/common/ftp_service.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class FTPService:
+    def __init__(
+        self,
+        host="localhost",
+        username="airflow",
+        password="airflow",
+        port=2222,
+        dir="/upload",
```
MJedr, line 19: maybe better to put it in config file (if you have one)?

dags/common/ftp_service.py
```diff
@@ -0,0 +1,79 @@
...
+            raise DirectoryNotFoundException(
+                "Remote directory doesn't exist. Abort connection."
+            )
+        return ftp
+
+    def __enter__(self):
+        self.connection = self.__connect()
+        return self
+
+    def __exit__(self, exc_type, exc_value, tb):
```
MJedr: can we call it `exception_type` and `exc_value` instead (if the naming is not forced by the library)?

dags/common/ftp_service.py
```diff
@@ -0,0 +1,79 @@
...
+        return True
+
+    def list_files(self, excluded_directories=None):
+        try:
+            file_names = []
+            filtered_files = []
+            walk_ftp(ftp=self.connection, remotedir=self.dir, paths=file_names)
+            for file_name in file_names:
+                append_file_if_not_in_excluded_directory(
+                    re.sub(self.dir + "/", "", file_name),
```
MJedr, line 65: maybe assign it as a variable with a proper name, it will be easier to know what's happening

dags/common/ftp_service.py
```diff
@@ -0,0 +1,79 @@
...
+                )
+            return filtered_files
+        except AttributeError:
+            raise NotConnectedException
+
+    def get_file(self, file):
+        try:
+            file_contents = io.BytesIO()
+            self.connection.retrbinary(
+                f"RETR {os.path.join(self.dir, file)}", file_contents.write
```
MJedr: again, maybe take out `os.path.join(self.dir, file)` from f-string and assign it to the variable

dags/common/ftp_service.py
ErnestaP: we don't have one, but I can write one :D

ErnestaP: @drjova I found an easy way to add configuration, by using .env: [dynaconf](https://dynaconf.readthedocs.io/en/docs_223/) 
With this tool, we don't need to load .env every time we want to reach values, it will be automatically loaded to the settings object.
You can find an example here, in _Option4: Dynaconf — Powerful settings configuration for Python_: https://towardsdatascience.com/from-novice-to-expert-how-to-write-a-configuration-file-in-python-273e171a8eb3

MJedr (approved):

dags/common/ftp_service.py
drjova: `dynaconf` it's a great tool. Integrating would need changes in the current infrastructure (k8s), and I would suggest keeping reading the env vars for now.

Furthermore, `airflow` seems that provide similar functionality with `Variables` https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/variables.html that you can have it as configuration JSON per environment, which is similar to `dynaconf`.

ErnestaP: If we want to use the Variables from airflow uploaded by JSON file, we will need to change k8s as well, by importing the file in airflow: `airflow variables import properties.json`
I will put in the make file, for development

PR #132, opened 2023-05-10
Integration tests fix

ErnestaP:
* Updated airflow.cfg file
* Updated Makefile, however, the celery worker is still requires manual starting and stopping
* Udded test for Hindawi dag triggering
* Fixed test: un-paused dag on dag run
*


tests/integration/oup/test_oup_pull_ftp.py
```diff
@@ -21,48 +21,61 @@ def dag():
...
+            )
+            assert len(repo.find_all()) == 3
 
-def get_dagrun_state(dagrun: DagRun):
-    dagrun.update_state()
-    return (
-        not dagrun.get_state() == DagRunState.QUEUED
-        and not dagrun.get_state() == DagRunState.RUNNING
-    )
+    def get_dagrun_state(self, dagrun: DagRun):
```
MJedr: maybe you can have this function in test utils instead of repeating it in all the classes?

tests/integration/aps/test_utils.py
```diff
@@ -23,49 +23,56 @@ def dag():
     return dagbag.get_dag(dag_id=DAG_NAME)
 
 
-def test_dag_loaded(dag: DAG):
-    assert dag is not None
-    assert len(dag.tasks) == 3
-
+@pytest.fixture
+def dag_was_paused(dag):
```
MJedr: again, better move it into one file and reuse it in other tests

tests/integration/springer/test_dag_process_file.py
```diff
@@ -47,39 +52,56 @@ def extract_zip_to_article(zip_filename):
...
-    springer_parse_file(
-        params={"file": base64.b64encode(ET.tostring(article)).decode()}
-    )
+class TestClassSpringerFilesHarvesting:
+    def test_dag_loaded(self, dag: DAG):
+        assert dag is not None
+        assert len(dag.tasks) == 4
+
+    def test_dag_run(self, dag: DAG, dag_was_paused: bool, article: ET):
+        id = datetime.datetime.utcnow().strftime(
```
MJedr: better not to call the vars in Python `id`, check https://stackoverflow.com/questions/77552/id-is-a-bad-variable-name-in-python

tests/integration/springer/test_dag_process_file.py
```diff
@@ -47,39 +52,56 @@ def extract_zip_to_article(zip_filename):
...
+    def test_dag_parse_file(self, article):
+        springer_parse_file(
+            params={"file": base64.b64encode(ET.tostring(article)).decode()}
+        )
+
+    def __test_dagrun_fails(self, dagrun):
+        dagrun.update_state()
+        return not dagrun.get_state() == DagRunState.FAILED
+
+    def __test_dagrun_state(self, dagrun):
```
MJedr: As I wrote, I would move it to another module. Also, it's a bit confusing for me: can a dag have more than 1 state? So if it's running for example it's not possible that it's failed or queued? If so, maybe you can have one assertion function where you can pass the desired state and assert it?

ErnestaP: No, the DAG **run** cannot have more than one state at a time

tests/integration/aps/test_utils.py
ErnestaP: I need to use it as a fixture because I need to preserve the initial state, since I am unpausing during the test (if the dag was paused) and then returning the dag for initial state as it was before the running the tests

MJedr: but what's the difference if you move it to a global conftest or you have it there?

MJedr (approved):

PR #133, opened 2023-05-16
Poetry

ErnestaP:


.github/workflows/test-and-build.yml
```diff
@@ -32,18 +32,27 @@ jobs:
...
+
+      - name: Start airflow services
+        run: |
+          poetry run airflow db init
+          poetry run airflow webserver -D
+          poetry run airflow triggerer -D
+          poetry run airflow scheduler -D
+          poetry run airflow celery worker -D
+          poetry run airflow celery flower -D
+          poetry run echo -e "\033[0;32m Airflow Started. \033[0m"
```
drjova: `make airflow`

Dockerfile
```diff
@@ -1,14 +1,18 @@
...
 COPY dags ./dags
-USER airflow
 RUN pip install --upgrade pip &&\
     pip install --no-cache-dir --upgrade setuptools==59.1.1 &&\
     pip install --no-cache-dir --upgrade wheel &&\
-    pip install --no-cache-dir --user -r requirements.txt -r requirements-test.txt
-    
+    pip install --no-cache-dir poetry
+
+RUN poetry install
```
drjova, line 18: hmm why we need to change completely the image?

Makefile
```diff
@@ -43,6 +43,7 @@ create_user:
 		--lastname LAST_NAME \
 		--email admin@worfklows.cern
 
+
```
drjova: no need :)

dags/common/cleanup.py
```diff
@@ -23,8 +23,8 @@ def clean_collaboration(input):
 
 
 def remove_specific_tags(value, tags=None, attributes=None):
-    tags = tags or []
-    attributes = attributes or []
+    tags = tags or {}
+    attributes = attributes or {}
```
drjova, line 27: any particular reason to use set instead of array?

requirements-airflow.txt
```diff
@@ -1,5 +1 @@
-apache-airflow[celery]==2.6.0 --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.6.0/constraints-3.8.txt
-apache-airflow-providers-celery==2.1.3
-celery==5.2.6
-redis==4.2.2
-psycopg2==2.9.6
+apache-airflow[celery,postgres,redis]==2.6.0 --constraint https://raw.githubusercontent.com/apache/airflow/constraints-2.6.0/constraints-3.10.txt
```
drjova: do we need to keep the requirements files since we move to poerty?

tests/units/common/test_cleanup.py
```diff
@@ -97,9 +97,9 @@ def test_convert_html_subsripts_to_latex(test_input, expected):
 
 
 xml = "<div><p>example<h1> h1 example</h1></p></div>"
-xml_just_p = "<p>example h1 example</p>"
-xml_just_div = "<div>example h1 example</div>"
-xml_div_and_h1 = "<div>example<h1> h1 example</h1></div>"
+xml_just_p = "<p>example\n h1 example</p>"
+xml_just_div = "<div>\nexample\n h1 example</div>"
+xml_div_and_h1 = "<div>\nexample<h1> h1 example</h1></div>"
```
drjova, line 102: hmm are these changes related to poetry?

ErnestaP: kind of yes. The upgrading airflow also upgrades the dependencies. As a result, the version of `bleach` upgraded as well, and it works slightly different than the older one

dags/common/cleanup.py
ErnestaP: It's a change in bleach lib: https://bleach.readthedocs.io/en/latest/changes.html

Dockerfile
ErnestaP: But we want to use poetry but not pip anymore. Instead of it, we had pip before:
`pip install --no-cache-dir --user -r requirements.txt -r requirements-test.txt`

requirements-airflow.txt
ErnestaP: no, we don't

.github/workflows/test-and-build.yml
```diff
@@ -32,18 +32,21 @@ jobs:
...
       - name: Create S3 Buckets
         run: make buckets
 
       - name: Run services for IT Tests
-        run: make start
+        run: make compose
+
+      - name: Start airflow services
+        run: |
+          poetry run make start
```
drjova: why we need poetry run make? 😕

.github/workflows/test-and-build.yml
```diff
@@ -32,18 +32,21 @@ jobs:
...
+          python -m pip install --upgrade pip
+          pip install poetry
+          poetry install
 
       - name: Create S3 Buckets
         run: make buckets
 
       - name: Run services for IT Tests
-        run: make start
+        run: make compose
```
drjova: `compose` is already included in the `make start` https://github.com/cern-sis/workflows/blob/main/Makefile#L17

Dockerfile
drjova: @ErnestaP i'm referring to the docker image that we are using here, it's changed from

```
FROM apache/airflow:2.6.0-python3.10
```

to 


```
FROM python:3.10
```

Is it any particular reason?

README.md
```diff
@@ -21,40 +21,52 @@ Currently the docker-compose performance isn't very good. Thus, for local dev, i
...
 A Makefile has been created to ease this process. The available targets are the following :
 
-- `make init` : Inits the pyenv version and virtual env.
-- `make start` : Starts all processes needed.
-- `make stop` : Stops all processes needed.
-- `make create_user` : create a new user with all the parameters
+- `poetry run make init` : Inits the pyenv version and virtual env.
+- `poetry run make start` : Starts all processes needed.
+- `poetry run make stop` : Stops all processes needed.
+- `poetry run make create_user` : create a new user with all the parameters
```
drjova: `poetry` and `make` are two independent tools, I guess what we want to achieve here is running with poetry the commands **inside** make, hence we should update `Makefile` instead. :)

pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[tool.poetry]
+name = "workflows"
+version = "0.1.0"
+description = ""
+authors = ["ErnestaP <ernesta.petraityte@yahoo.com>"]
```
drjova: `cern-sis <cern-sis-ts-admin@cern.ch>`

tests/units/common/test_cleanup.py
drjova: Thanks for the explanation, but still I don't understand why the new lines. Do we have to configure it differently?

drjova (changes requested):

.github/workflows/test-and-build.yml
ErnestaP: I will move "poetry run ..." inside make file

README.md
ErnestaP: Dockerfile - yes. The reason is I want to see if airflow runs by using poetry. That's why I am taking the Python image, installing poetry, and then running tests by using this container. 

https://github.com/cern-sis/workflows/blob/9a0124e82d48c505f910783547daa7ec6dc8532c/.github/workflows/test-and-build.yml#L72-L76 (the file is from Poetry PR)

.github/workflows/test-and-build.yml
drjova: please read my comment https://github.com/cern-sis/workflows/pull/133#discussion_r1266501345

tests/units/common/test_cleanup.py
ErnestaP: No, there was no changes made from our side for bleach

drjova: Yes I understand that we didn't do any changes and it comes from bleach. What I'm saying is, if we should configure bleach differently, since it seems wrong to have new lines.

ErnestaP: I don't like it as well. The new line appears instead of the cleaned starting tag, when we are cleaning the tags which are not surrounded. Let's say we have:
`<div><p></p></div>`
we clean the string from "p", the result will be:
`<div>\n</div>`

Makefile
```diff
@@ -8,11 +8,10 @@ WORKER_PID=airflow-worker.pid
 FLOWER_PID=airflow-flower.pid
 
 init:
-	pyenv global $(PYTHON_VERSION)
-	pyenv install ${PYTHON_VERSION}
-	pyenv virtualenv ${PYTHON_VERSION} workflows
-	pyenv activate workflows
+	poetry env use 3.11
```
drjova: with pyenv we ensure that the right python version is installed and enabled

Makefile
```diff
@@ -8,11 +8,10 @@ WORKER_PID=airflow-worker.pid
...
+	poetry shell
```
drjova: why do we need `poetry shell`? if we run poetry shell then there is no need to run `poetry run`

README.md
```diff
@@ -21,31 +21,43 @@ Currently the docker-compose performance isn't very good. Thus, for local dev, i
...
-1. First, we should create virtual environment with pyenv:
+1. First, we need to install poetry:
 
-```
-    export PYTHON_VERSION=3.8.13
-    pyenv install $PYTHON_VERSION
-    pyenv global $PYTHON_VERSION
-    pyenv virtualenv $PYTHON_VERSION workflows
-    pyenv activate workflows
-```
```
drjova, line 32: we should keep the python version installation in the documentation

README.md
```diff
@@ -21,31 +21,43 @@ Currently the docker-compose performance isn't very good. Thus, for local dev, i
...
-5. Run Airflow. Airflow comes with the `standalone` command. This should not be used as it seems to use the SequentialExecutor by default. To have our local config being use, we must run all services together :
+3. Set airflow home directory:
 
-```
-    airflow webserver
-    airflow triggerer
-    airflow scheduler
-```
+    ```
+    export AIRFLOW_HOME=/path/to/cloned/repo
```
drjova:
```suggestion
    export AIRFLOW_HOME=${PWD}
```

README.md
```diff
@@ -21,31 +21,43 @@ Currently the docker-compose performance isn't very good. Thus, for local dev, i
...
+    If you are using incorrect python version, you will get an error:
+    ```
+    current Python version (X.Y.Z) is not allowed by the project (^X.Y).
+    Please change python executable via the "env use" command.
+    ```
+    You can easily fix it with the following command:
+    ```
+    poetry env use X.Y
+    ```
+    The correct python version can be found in pyptoject.toml file
```
drjova: hmm is this copy paste from somewhere else? we should specify the version that the project uses.

tests/integration/aps/test_utils.py
```diff
@@ -50,7 +50,6 @@ def test_dag_run(dag: DAG):
     repo = APSRepository()
     repo.delete_all()
     assert len(repo.find_all()) == 0
-    dag.clear()
```
drjova: why we have to remove it?

tests/units/common/test_cleanup.py
drjova: We have to understand the impact of this change, we have quite a few functions in cleanup that uses bleach.

Dockerfile
drjova: @ErnestaP ☝️

drjova (changes requested):

Dockerfile
ErnestaP: Sorry, I replied in the wrong place... 

Dockerfile - yes. The reason is I want to see if airflow runs by using poetry. That's why I am taking the Python image, installing poetry, and then running tests by using this container. 

https://github.com/cern-sis/workflows/blob/9a0124e82d48c505f910783547daa7ec6dc8532c/.github/workflows/test-and-build.yml#L72-L76 (the file is from Poetry PR)

README.md
ErnestaP: No, I wrote by myself. The python version might change with every update, am I right?

tests/integration/aps/test_utils.py
ErnestaP: it was crashing on Github actions. Cannot remember the exact error, but the github action log just froze, if I wanted to see the traceback.

tests/units/common/test_cleanup.py
ErnestaP: Should we try to downgrade the bleach version? Because I spent a day in order to understand what actually changed, what we are getting the newlines introduced. However, I wasn't able to find something about this in the documentation

tests/integration/aps/test_utils.py
drjova: crashing github actions? This should not be related maybe there was an outage. We shouldn't just delete things without understanding the issue :)

.github/workflows/test-and-build.yml
```diff
@@ -32,17 +32,13 @@ jobs:
...
-          pip install --upgrade pip
-          pip install --no-cache-dir --upgrade setuptools
-          pip install --no-cache-dir --upgrade wheel
-          pip install -r requirements.txt -r requirements-test.txt -r requirements-airflow.txt
+          python -m pip install --upgrade pip
+          pip install poetry
+          poetry install
 
-      - name: Create S3 Buckets
-        run: make buckets
```
drjova, line 43: [make start](https://github.com/cern-sis/workflows/blob/main/Makefile#L17) doesn't include buckets

README.md
drjova: if we change the python version we should update the docs, let's keep the version that we would like to run it

Makefile
drjova: Also, why 3.11? in dockerfile we are using 3.10 :)

.github/workflows/test-and-build.yml
ErnestaP: But compose do

tests/units/common/test_cleanup.py
ErnestaP: I cannot find anything on the change log regarding the new lines.
They already appear on 5.0.0 version 
https://bleach.readthedocs.io/en/latest/changes.html#version-5-0-0-april-7th-2022

The PR which is related to clean function, should not make a difference. https://github.com/mozilla/bleach/pull/644/files

tests/integration/aps/test_utils.py
ErnestaP: M... put back `dag.clean()`, the error haven't been raised again

tests/units/common/test_cleanup.py
ErnestaP: So, what would be a better option:
1. Downgrade to 4.0.1 (now bleach version is 6.0.0)
2. Keep the newest versions as a cost with new lines (as it is now) in our tests

README.md
```diff
@@ -21,31 +21,39 @@ Currently the docker-compose performance isn't very good. Thus, for local dev, i
...
+
+   ```
+   curl -sSL https://install.python-poetry.org | python3 -
+
+   ```
+
+2. Install pyenv, the instruction can be found [here](https://github.com/pyenv/pyenv#installation)
+3. Export correct python version, create and activate virtual environment by using pyenv:
+   ```
+   export PYTHON_VERSION = 3.8.13
```
drjova: why 3.8.13? in docker we are using 3.10 :)

pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+inspire-utils = "3.0.25"
+furl = "^2.1.3"
+pre-commit = "3.2.2"
+pytest = "7.1.0"
+coverage = "6.3.2"
+pytest-cov = "^4.0.0"
+pytest-vcr = "1.0.2"
+pytest-datadir = "1.3.1"
+freezegun = "1.2.1"
+requests-mock = "1.9.3"
```
drjova: I guess all these should be in the`--dev` deps

README.md
```diff
@@ -21,31 +21,39 @@ Currently the docker-compose performance isn't very good. Thus, for local dev, i
...
+   pyenv global $(PYTHON_VERSION)
+   pyenv install ${PYTHON_VERSION}
```
drjova: the other way around :)

README.md
```diff
@@ -21,31 +21,39 @@ Currently the docker-compose performance isn't very good. Thus, for local dev, i
...
+   pyenv virtualenv ${PYTHON_VERSION} workflows
```
drjova: no need this because we are using poetry which comes with each own vm and we are modifying all the commands to use it `poetry run ...`

Makefile
```diff
@@ -7,13 +6,6 @@ SCHEDULER_PID=airflow-scheduler.pid
 WORKER_PID=airflow-worker.pid
 FLOWER_PID=airflow-flower.pid
 
-init:
-	pyenv global $(PYTHON_VERSION)
-	pyenv install ${PYTHON_VERSION}
-	pyenv virtualenv ${PYTHON_VERSION} workflows
-	pyenv activate workflows
-	export AIRFLOW_HOME=${PWD}
```
drjova: Why removing it? We just need to update it and to align it with the documentation

pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[tool.poetry]
+name = "workflows"
+version = "0.1.0"
+description = ""
+authors = ["cern-sis <cern-sis-ts-admin@cern.ch>"]
+readme = "README.md"
+
+[tool.poetry.dependencies]
+python = "^3.10"
+apache-airflow = {version = "2.6.0", extras = ["celery", "postgres", "redis"]}
```
drjova, line 10: isn't it the `2.6.1` the latest?

ErnestaP: it is, but when we started this PR it was 2.6.0. Let's change to 2.6.1

Makefile
ErnestaP: I understood wrong: we should keep the python version installation in the documentation
That we should keep just in documentation

PR #138, opened 2023-05-26
Workflows: tests fixes

ErnestaP:
* Use dag.test() in order to quickly test the whole dag run

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,11 +1,11 @@
-export PYTHON_VERSION = 3.8.13
+export PYTHON_VERSION = 3.10.11
 
-
-WEBSERVER_PID=airflow-webserver-monitor.pid
-TRIGGERER_PID=airflow-triggerer.pid
-SCHEDULER_PID=airflow-scheduler.pid
-WORKER_PID=airflow-worker.pid
 FLOWER_PID=airflow-flower.pid
+SCHEDULER_PID=airflow-scheduler.pid
+TRIGGERER_PID=airflow-triggerer.pid
+WEBSERVER_MONITOR_PID=airflow-webserver-monitor.pid
+WEBSERVER_PID=airflow-webserver.pid
+WORKER_PID=airflow-worker.pid
 
 init:
 	pyenv global $(PYTHON_VERSION)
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,14 +1,11 @@
-FROM apache/airflow:2.6.0-python3.10
-
-ENV PYTHONBUFFERED=0 
-ENV AIRFLOW_UID=501
+FROM apache/airflow:2.6.0-python3.8
```
drjova: why are we using python `3.8`?

dags/oup/ftp_service.py
```diff
@@ -5,10 +5,6 @@
...
     def __init__(self):
-        super().__init__(
-            host=os.getenv("OUP_FTP_HOST", "localhost"),
-            username=os.getenv("OUP_FTP_USERNAME", "airflow"),
-            password=os.getenv("OUP_FTP_PASSWORD", "airflow"),
-            port=int(os.getenv("OUP_FTP_PORT", "2222")),
-            dir=os.getenv("OUP_FTP_DIR", "upload/oup"),
-        )
+        super().__init__()
+        self.ftp_conn_id = "ftp_service"
```
drjova: we can make it configurable? and have a default name something like `oup_ft_service`?

dags/oup/oup_pull_ftp.py
```diff
@@ -21,7 +20,7 @@ def oup_pull_ftp():
 
     @task()
     def migrate_from_ftp(
-        ftp: FTPService = OUPFTPService(), repo: IRepository = OUPRepository(), **kwargs
+        ftp=OUPFTPService(), repo: IRepository = OUPRepository(), **kwargs
```
drjova: why deciding to drop the types, we should keep it :)

tests/integration/iop/test_dag_pull_sftp.py
```diff
@@ -77,13 +77,5 @@ def test_pull_from_sftp(iop_empty_repo):
             },
             {"xml": "extracted/aca95c.xml/aca95c.xml"},
         ]
+
         assert iop_empty_repo.find_all() == expected_files
-        assert sorted(iop_empty_repo.get_all_raw_filenames()) == sorted(
```
drjova: in some tests we keep this in some others we don't. I would suggest to keep them in all ☺️

Dockerfile
ErnestaP: Because I aimed to keep the same Python version, as in previous tests, to be sure that it is not the version's fault. 
I can change it to 3.10 (3.11 is crashing on my computer, while trying to install dependencies), since it is obviously not the version fault

PR #146, opened 2023-06-23
Workflows: tests fixes

ErnestaP:
* Use dag.test() in order to quickly test the whole dag run
* Separated dag and harvesting methods tests


dags/common/utils.py
```diff
@@ -198,3 +200,5 @@ def check_dagrun_state(dagrun: DagRun, not_allowed_states=[], allowed_states=[])
         value = dagrun.get_state() == dag_run_states[allowed_state]
         states_values.append(value)
     return all(states_values)
+def generate_ids():
+    return "".join(random.choices(string.ascii_uppercase + string.digits, k=10))
```
MJedr: I think it's better to use uuid module https://docs.python.org/2/library/uuid.html#module-uuid

tests/integration/iop/test_dag_pull_sftp.py
```diff
@@ -77,13 +77,7 @@ def test_pull_from_sftp(iop_empty_repo):
...
-                "2022-09-01T03_01_40_content.zip",
-                "2022-09-03T03_01_49_content.zip",
-                "2022-09-24T03_01_43_content.zip",
-                "aca95c.xml.zip",
-            ]
-        )
+
+        assert sorted(
+            iop_empty_repo.find_all(), key=lambda x: x["xml"], reverse=True
+        ) == sorted(expected_files, key=lambda x: x["xml"], reverse=True)
```
MJedr, line 83: maybe use set equality for that, it will be more efficient

MJedr (commented):
It's much cleaner now!

dags/common/utils.py
ErnestaP: Is it better to install the modules, if I am using it just in one place? Or is it better for readability?

tests/integration/iop/test_dag_pull_sftp.py
ErnestaP: since it's a list, do you have in mind, iterate per each element, use set, and then compare?:

like: 
`[for set(files_from_repo) in iop_empty_repo.find_all()] == [for set(expected_file) in expected_files]`

MJedr: no, why? I was thinking set(iop_empty_repo.find_all()) instead of sorted

dags/common/utils.py
MJedr: uuid is built in in python

tests/integration/iop/test_dag_pull_sftp.py
ErnestaP: because, we can put list in the set 😔 : 
`TypeError: unhashable type: 'dict'`

ErnestaP: because we cannot put list in the set 😔, we will get an error:
`TypeError: unhashable type: 'dict'`

MJedr: ok, it's not a big deal :)

PR #152, opened 2023-07-17
enhancer: removed country from records

ErnestaP:
* Removed country from records with affiliation value with: cooperation agreement with cern.
* ref: https://github.com/cern-sis/issues-scoap3/issues/176

### DIFF
--- a/dags/common/enhancer.py
+++ b/dags/common/enhancer.py
@@ -1,4 +1,5 @@
 import datetime
+import re
 
 from common.constants import FN_REGEX
 
@@ -42,6 +43,18 @@ class Enhancer:
             }
         ]
 
+    def __remove_country(self, item):
+        pattern_for_cern_cooperation_agreement = re.compile(
+            r"cooperation agreement with cern", re.IGNORECASE
+        )
+        for author in item.get("authors", []):
+            for affiliation in author.get("affiliations", []):
+                match_pattern = pattern_for_cern_cooperation_agreement.search(
+                    affiliation.get("value", "")
+                )
+                if match_pattern:
+                    affiliation.pop("country", None)
+
     def __call__(self, publisher, item):
         creation_date = datetime.datetime.now().isoformat()
         item_copy = item.copy()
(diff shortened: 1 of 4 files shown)


dags/common/enhancer.py
```diff
@@ -42,6 +43,18 @@ def __construct_titles(self, item, publisher):
             }
         ]
 
+    def __remove_country(self, item):
+        for author in item.get("authors", []):
+            for affiliation in author.get("affiliations", []):
+                pattern_for_cern_cooperation_agreement = re.compile(
+                    r"cooperation agreement with cern", re.IGNORECASE
```
drjova: if we have the `compile` inside the loops it's loosing its advantage over the simple `match`, especially in a double for loop.

drjova (changes requested):
Thank you 😃 , there is a comment that we have to address.

tests/units/common/test_enhancer.py
```diff
@@ -60,6 +60,128 @@
...
+    "subtitle": "subtitle",
+}
+expected_output_with_affiliation_of_cooperation_agreement_with_CERN = {
+    "abstracts": [{"value": "this is abstracts", "source": publisher}],
+    "authors": [
+        {
+            "affiliations": [
+                {
+                    "organization": "International laboratory covered by a cooperation agreement with CERN",
+                    "value": "International laboratory covered by a cooperation agreement with CERN, Geneva, Switzerland",
```
drjova, line 121: we are not testing the case that `country` already populated from the XML files.

dags/common/enhancer.py
drjova: @ErnestaP ☝️

drjova (approved):

drjova:
@ErnestaP could you please squash?

drjova:
@ErnestaP why there is a merge commit?

ErnestaP:
Merge commit because of resolving the merge conflict by using UI

PR #157, opened 2023-07-26
Revert poetry changes back to PIP

ErnestaP:


tests/integration/aps/test_utils.py
```diff
@@ -41,20 +46,19 @@ def test_aps_fetch_api(dag: DAG):
...
     dag.test(
         run_conf={
-            "from_date": "2022-02-05",
-            "until_date": "2022-03-05",
+            "from_date": "2022-02-07",
+            "until_date": "2022-02-07",
         }
     )
-    assert len(repo.find_all()) == 1
+    time.sleep(5)
```
ErnestaP, line 63: the tests on github actions are unstable. You can see that fetch files from the FTP server were not saved before the  `assert len(aps_empty_repo.find_all()) == 1` was checked. So, the file was not found in our s3 pod.
https://github.com/cern-sis/workflows/actions/runs/5679947342/job/15393077928

This test was always passing before, however, now we see that it's crashing. I put `sleep` in order to avoid this issue. But it's not a good solution. Any suggestions?

tests/integration/aps/cassettes/test_dag_run.yaml
ErnestaP, line 1: I thought a smaller file will make a difference in test passing. Unfortunately, it didn’t.
The test test_dag_run in test/itegration/aps/test_utils.py passes locally, but not on gh actions.
Also, sometime it passes, sometimes it not

ErnestaP: any ideas, why this test is crashing?

drjova (approved):

PR #168, opened 2023-09-26
Documentation: Hindawi

ErnestaP:


dags/hindawi/parser.py
```diff
@@ -133,7 +133,7 @@ def _get_arxiv(self, article: ET.Element):
             self.prefixes,
         )
         if not arxivs:
-            self.logger.error("No arxiv id_get_copyright_statement found.")
+            self.logger.error("No arxiv _get_arxiv found.")
```
drjova: what that means? :) Should it be no arxiv id found?

drjova (approved):

PR #169, opened 2023-10-02
Elsevier harvester

ErnestaP:


dags/common/pull_ftp.py
```diff
@@ -21,21 +22,26 @@ def migrate_files(
...
-            logger.msg("File is not a zipfile, processing next file.")
-            continue
-        with zipfile.ZipFile(file_bytes) as zip:
-            for zip_filename in zip.namelist():
-                file_prefix = ".".join(_file.split(".")[:-1])
-                file_content = zip.read(zip_filename)
-                s3_filename = os.path.join(file_prefix, zip_filename)
-                repo.save(s3_filename, io.BytesIO(file_content))
+
+    for file_name in filenames:
```
MJedr: if it's a zipfile/tarfile maybe you can change variable naming (`tar_file` instead of `file_name`) cause it's a bit confusing (at least for me)

dags/common/utils.py
```diff
@@ -198,3 +201,30 @@ def check_dagrun_state(dagrun: DagRun, not_allowed_states=[], allowed_states=[])
         value = dagrun.get_state() == dag_run_states[allowed_state]
         states_values.append(value)
     return all(states_values)
+
+
+def process_zip_file(file_bytes, file_name):
+    file_bytes.seek(0)
```
MJedr, line 207: Why do you need set the pointer explicitly? Is there any other place where you read the data?

dags/common/utils.py
```diff
@@ -198,3 +201,30 @@ def check_dagrun_state(dagrun: DagRun, not_allowed_states=[], allowed_states=[])
...
+    with zipfile.ZipFile(file_bytes) as zip:
+        for zip_filename in zip.namelist():
+            file_prefix = ".".join(file_name.split(".")[:-1])
+            zip_file_content = zip.read(zip_filename)
+            s3_filename = os.path.join(file_prefix, zip_filename)
+            yield (zip_file_content, s3_filename)
```
MJedr, line 213: Again, for me the names are confusing cause afaiu `zip_filename` is one of the files packed in zip file. So in loop this is not actually `zip_filename` but normal file name?

dags/common/utils.py
```diff
@@ -198,3 +201,30 @@ def check_dagrun_state(dagrun: DagRun, not_allowed_states=[], allowed_states=[])
...
+
+
+def process_tar_file(file_bytes, file_name):
+    file_bytes.seek(0)
+    with tarfile.open(fileobj=file_bytes, mode="r") as tar:
+        for tar_filename in tar.getnames():
+            file_prefix = ".".join(file_name.split(".")[:-1])
+            tar_file_content = tar.extractfile(tar_filename).read()
```
MJedr: same as before

dags/elsevier/elsevier_pull_ftp.py
```diff
@@ -0,0 +1,53 @@
...
+            "filenames_pull" in params
+            and params["filenames_pull"]["enabled"]
+            and not params["filenames_pull"]["force_from_ftp"]
+        )
+        if specific_files:
+            specific_files_names = pull_ftp.reprocess_files(repo, logger, **kwargs)
+            return specific_files_names
+
+        with ftp:
+            return pull_ftp.migrate_from_ftp(ftp, repo, logger, **kwargs)
```
MJedr, line 39: TBH If I was writing this dag, I would branch task `migrate_from_ftp` and then depending on the kwargs delegate `pull_ftp.reprocess_file` or `pull_ftp.migrate_from_ftp`. The reason is that this is totally different kind of processing in the end and it would be good to keep track what's actually happening in the dag run. But you can ignore the comment, it's rather to discuss the options :)

dags/elsevier/repository.py
```diff
@@ -0,0 +1,61 @@
...
+    ZIPED_DIR: str = "raw/"
+    EXTRACTED_DIR: str = "extracted/"
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.s3 = S3Service(os.getenv("ELSEVIER_BUCKET_NAME", "elsevier"))
+
+    def get_all_raw_filenames(self):
+        return [
+            f.key.replace("raw/", "")
```
MJedr: I think `lstrip` might be better in this case

dags/elsevier/repository.py
```diff
@@ -0,0 +1,61 @@
...
+            for f in self.s3.objects.filter(Prefix=self.ZIPED_DIR).all()
+        ]
+
+    def find_all(self, filenames_to_process=None):
+        grouped_files = {}
```
MJedr, line 25: I think you need a better name for this dict (might be longer but more meaningful), it's confusing

dags/elsevier/repository.py
```diff
@@ -0,0 +1,61 @@
...
+            filenames_to_process
+            if filenames_to_process
+            else self.__find_all_extracted_files()
+        )
+        for file in filenames:
+            last_part = os.path.basename(file)
+            filename_without_extension = last_part.split(".")[0]
+            if filename_without_extension not in grouped_files:
+                grouped_files[filename_without_extension] = {}
+            extension = find_extension(last_part)
```
MJedr: it can raise exception that you're not handling, it will break the loop then

dags/elsevier/repository.py
```diff
@@ -0,0 +1,61 @@
...
+            grouped_files[filename_without_extension][extension] = file
```
MJedr: I don't understand why you need the grouping, the whole function looks a bit overcomplicated to me

dags/elsevier/repository.py
```diff
@@ -0,0 +1,61 @@
...
+
+    def save(self, filename: str, obj: IO):
+        obj.seek(0)
+        prefix = self.ZIPED_DIR if ".tar" in filename else self.EXTRACTED_DIR
+        self.s3.upload_fileobj(obj, prefix + filename)
+
+    def delete_all(self):
+        self.s3.objects.all().delete()
+
+    def __find_all_extracted_files(self):
```
MJedr: why this method has to be private?

dags/elsevier/repository.py
```diff
@@ -0,0 +1,61 @@
...
+        return [
+            f.key
+            for f in self.s3.objects.filter(Prefix=self.EXTRACTED_DIR).all()
+            if self.is_meta(f.key) or ".pdf" in f.key
+        ]
+
+    def is_meta(self, filename: str):
```
MJedr: do you really need it? you use it once only and it's a bit inconsistent cause for the check `".pdf" in f.key` you don't have a special function

dags/common/pull_ftp.py
ErnestaP: mm.. it can be zip, tar, txt.. whatever file which is in FTP

dags/common/utils.py
ErnestaP: yes, I need to reset it before extracting it:
https://stackoverflow.com/a/75560385

dags/common/utils.py
ErnestaP: I agree, yes

dags/elsevier/elsevier_pull_ftp.py
ErnestaP: yes, would be nice to have a discussion about!

dags/elsevier/repository.py
ErnestaP: for this one, I might make a new PR. If I will change it for Eslevier, I would prefer also change in all other publishers

dags/elsevier/repository.py
ErnestaP: the idea of this function:
We can have files that have the same name but different extensions. 

A.xml.zip
A.pdf.zip
A.archival.zip

B.xml.zip
B.pdf.zip
B.archival.zip

```
[
  {
            "pdf": ""extracted/A/date_A.pdf",
            "xml": ""extracted/A/date_A.xml",
            "pdfa": "extracted/A/date_A.pdfa"
        },
  {
            "pdf": ""extracted/B/date_B.pdf",
            "xml": ""extracted/B/date_B.xml",
            "pdfa": "extracted/B/date_B.pdfa"
        }
]

```

dags/elsevier/repository.py
ErnestaP: because it's inner function, it has to be used just in the [find_all](https://github.com/cern-sis/workflows/blob/4b054c906a2f81ae3a818d7a64d1656c996cdfaf/dags/elsevier/repository.py#L29) method:
https://github.com/cern-sis/workflows/blob/4b054c906a2f81ae3a818d7a64d1656c996cdfaf/dags/elsevier/repository.py#L29

dags/elsevier/repository.py
ErnestaP: yes, for Elsevier is not needed

dags/common/pull_ftp.py
MJedr: just an idea: can't you filter files that you pull from ftp so that you get files with desired extensions only? do they really put everything in one directory on ftp? 🤦‍♀️

dags/elsevier/repository.py
MJedr: ☝️ instead of replace, won't that work? you don't go through the whole string to find matching pattern ;)

dags/elsevier/repository.py
MJedr: it's not inner function in python terms ;) inner function is sth like
```
def outer():
    a = []
    def inner():
        a.append(5)
    return inner

```
you can make protected maybe, it's very rare to use private method's in python (unlike in Java, for example). IMO if you go into this direction you should be very consistent about it.

dags/common/pull_ftp.py
ErnestaP: Sorry, it was wrong what I said. Actually it can be zip or tar, that's it. ZIP for IOP, OUP, Springer, TAR for Elsevier

ErnestaP: but inside of archive, we have xml, txt files.
I don't think we can call tar_file maybe `archive_file` is better?

dags/elsevier/repository.py
ErnestaP: It removes "r" , "a", "w", "/" from a string, which is not wanted behavior.
Screenshot 2023-10-09 at 08 18 14

I used  removeprefix instead:
Screenshot 2023-10-09 at 08 23 17

MJedr (approved):
looks good to me (not a big fan of the `find_all` method though ;))
don't forget to squash commits

PR #170, opened 2023-10-17
Elsevier parsing

ErnestaP:


dags/elsevier/metadata_parser.py
```diff
@@ -0,0 +1,119 @@
...
+                extraction_function=self._get_collections,
+                # required=True,
+            ),
+            CustomExtractor(
+                destination="license",
+                extraction_function=self._get_license,
+                # required=True,
+            ),
+            # CustomExtractor(
+            #     destination="local_files",
```
ErnestaP, line 44: should we keep it? It might be useful in the future. This field is in the current hepcrawl, but it doesn't pass the validation since it is not in the [schema](https://repo.qa.scoap3.org/schemas/hep.json). 
In schema we have _files:
```
 "_files": {
            "title": "File schema",
            "type": "array",
            "properties": {
                "key": {
                    "minLength": 1,
                    "type": "string"
                },
                "checksum": {
                    "minLength": 1,
                    "type": "string"
                },
                "size": {
                    "type": "integer"
                },
                "bucket": {
                    "minLength": 1,
                    "type": "string"
                },
                "version_id": {
                    "minLength": 1,
                    "type": "string"
                },
                "filetype": {
                    "minLength": 1,
                    "type": "string"
                }
            }
        },
```

tests/units/elsevier/test_metadata_parser.py
```diff
@@ -0,0 +1,72 @@
...
+            [
+                {
+                    "license": "CC-BY-3.0",
+                    "url": "http://creativecommons.org/licenses/by/3.0/",
+                }
+            ],
+            "license",
+            id="test_license",
+        ),
+        # param(
```
ErnestaP, line 53: https://github.com/cern-sis/workflows/pull/170/files#r1361673721

dags/elsevier/elsevier_file_processing.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+def enrich_elsevier(enhanced_file):
+    return Enricher()(enhanced_file)
+
+
+def elsevier_parse_metadata(enriched_file, repo):
+    file_path = enriched_file["source_file_path"]
+    dataset_file_path = file_path.split("/")[:-2]
+    file = repo.get_by_id("/".join(dataset_file_path + ["dataset.xml"]))
```
MJedr: I remember that for managing paths you were using pathlib/os at some files, maybe it's nice to stick to one way?

dags/elsevier/metadata_parser.py
```diff
@@ -0,0 +1,113 @@
...
+            #     # required=True,
+            # ),
+        ]
+
+    def parse(self, article):
+        extracted_value = {}
+        journal_issues = article.findall("dataset-content/journal-item")
+        for journal_issue in journal_issues:
+            doi = journal_issue.find("journal-item-unique-ids/doi")
+            if doi.text == self.dois:
```
MJedr, line 55: why this check?

dags/elsevier/metadata_parser.py
```diff
@@ -0,0 +1,113 @@
...
+                self.file_path, article.find("files-info/web-pdf/pathname").text
+            ),
+        }
+        xml = {
+            "filetype": "xml",
+            "path": os.path.join(
+                self.file_path, article.find("files-info/ml/pathname").text
+            ),
+        }
+        return [pdf, xml]
```
MJedr, line 113: if the order matters I would use dict to return it, just to be sure what you're accessing

dags/elsevier/elsevier_file_processing.py
```diff
@@ -0,0 +1,82 @@
...
+    def enchance(parsed_file):
+        return parsed_file and enhance_elsevier(parsed_file)
+
+    @task()
+    def enrich(enhanced_file):
+        return enhanced_file and enrich_elsevier(enhanced_file)
+
+    @task()
+    def parse_metadata(enriched_file, repo: IRepository = ElsevierRepository()):
+        return enriched_file and elsevier_parse_metadata(enriched_file, repo)
```
MJedr: why you have to return sth **and** func(sth)? can't you do 
```
func(sth) 
return sth
```

dags/elsevier/metadata_parser.py
```diff
@@ -0,0 +1,113 @@
+import os
+from datetime import datetime
+
+from common.parsing.parser import IParser
+from common.parsing.xml_extractors import CustomExtractor
+from common.utils import extract_text, get_license_type_and_version_from_url
+from structlog import get_logger
+
+
+class ElsevierMetadataParser(IParser):
```
MJedr, line 10: why you have to have metadata parser that parses same file as parser?

dags/elsevier/metadata_parser.py
ErnestaP: Information is parsed from 2 files, for the same article: from actual article XML file and form dataset,xml file that stores data about journals, collections, and licenses. Dataset.xml has more than one article's metadata in the same file, and they have their own block with data.
For example:
```
			<journal-item-unique-ids>
				<pii>S0370-2693(23)00355-6</pii>
				<doi>10.1016/j.physletb.2023.138021</doi>
				<jid-aid>
					<jid>PLB</jid>
					<issn>0370-2693</issn>
					<aid>138021</aid>
					<article-number>138021</article-number>
				</jid-aid>
			</journal-item-unique-ids>
		.......
			<journal-item-unique-ids>
				<pii>S0370-2693(23)00378-7</pii>
				<doi>10.1016/j.physletb.2023.138044</doi>
				<jid-aid>
					<jid>PLB</jid>
					<issn>0370-2693</issn>
					<aid>138044</aid>
					<article-number>138044</article-number>
				</jid-aid>
			</journal-item-unique-ids>
```
So I want to find the exact block from where I have to take the data about journal, license and etc.

dags/elsevier/metadata_parser.py
ErnestaP: not the same, it parses different file

dags/elsevier/metadata_parser.py
ErnestaP: thank you! But no, order doesn't matter

dags/elsevier/elsevier_file_processing.py
ErnestaP: if enriched_file is not a None value, I am calling elsevier_parse_metadata. If the value is None, it will return None

MJedr: ok, so if I understand correctly it might happen that you get `None` from parse. But then maybe it would be worth to finish the dagrun instead of performing all the other tasks?

dags/elsevier/metadata_parser.py
MJedr: ok 👍

dags/elsevier/elsevier_file_processing.py
ErnestaP: good idea

MJedr (approved):

PR #179, opened 2023-11-06
Elsevier dataset parsing

ErnestaP:


dags/common/pull_ftp.py
```diff
@@ -66,10 +71,10 @@ def migrate_from_ftp(
     )
 
     if force_pull_all_files:
-        return _force_pull(s_ftp, repo, logger, **kwargs)
+        return _force_pull(s_ftp, repo, logger, publisher, **kwargs)
     elif force_pull_specific_files:
-        return _filenames_pull(s_ftp, repo, logger, **kwargs)
-    return _differential_pull(s_ftp, repo, logger, **kwargs)
+        return _filenames_pull(s_ftp, repo, logger, publisher, **kwargs)
+    return _differential_pull(s_ftp, repo, logger, publisher, **kwargs)
```
drjova, line 85: why having a flag for that? Can we just use a different function for elsevier?

dags/common/utils.py
```diff
@@ -228,3 +227,61 @@ def process_archive(file_bytes, file_name):
...
+
+def process_zip_file_elsevier(file_bytes, file_name):
+    file_bytes.seek(0)
+    with zipfile.ZipFile(file_bytes) as zip:
+        for filename in zip.namelist():
+            if "dataset.xml" in filename:
+                file_prefix = ".".join(file_name.split(".")[:-1])
+                zip_file_content = zip.read(filename)
+                s3_filename = os.path.join(file_prefix, filename)
+                return (zip_file_content, s3_filename)
```
drjova: I would use the following pattern in order to avoid nested branches :)

```suggestion
        for filename in zip.namelist():
            if "dataset.xml" not in filename:
                continue
            file_prefix = ".".join(file_name.split(".")[:-1])
            zip_file_content = zip.read(filename)
            s3_filename = os.path.join(file_prefix, filename)
            return (zip_file_content, s3_filename)
```

dags/elsevier/metadata_parser.py
```diff
@@ -40,26 +44,36 @@ def __init__(self, doi, file_path) -> None:
...
                 required=True,
             ),
-            # CustomExtractor(
-            #     destination="local_files",
-            #     extraction_function=self._get_local_files,
-            #     # required=True,
-            # ),
+            CustomExtractor(
+                destination="files",
+                extraction_function=self._get_local_files,
```
drjova, line 49: aren't we ignoring local files?

dags/elsevier/trigger_file_processing.py
```diff
@@ -0,0 +1,61 @@
...
+                    dag_id=f"{publisher}_process_file",
+                    run_id=_id,
+                    conf={
+                        "file": encoded_article,
+                        "file_name": full_file_path,
+                        "metadata": parsed_article,
+                    },
+                    replace_microseconds=False,
+                )
+                files.append(full_file_path)
```
drjova: Do we really need the three level for loop?

dags/common/pull_ftp.py
ErnestaP: we have a flag for it because we have different files migration flow for Elsevier.
The solution with a flag helps to avoid boilerplate code.

Files list comming from  _differential_pull, _filenames_pull or _force_pull are the same to migrate_files or migrate_files_elsevier:
https://github.com/cern-sis/workflows/blob/elsevier-dataset-parsing/dags/common/pull_ftp.py#L97-L99

However, the function of migrate_files and migrate_files_elsevier is different.
Migrate_files_elsevier saves just archives: https://github.com/cern-sis/workflows/blob/elsevier-dataset-parsing/dags/common/pull_ftp.py#L181-L209
While common Migrate_files saves and extracts archives:
https://github.com/cern-sis/workflows/blob/elsevier-dataset-parsing/dags/common/pull_ftp.py#L18-L51

For Elsevier publisher sends us too many files which we don't need. Extracting them takes a lot of time, so we want to extract just what we need.
Elsevier adds mapping in dataset.xml, from where we are figuring out which files we actually need to extract and parse.
The extracting only needed files for Elsevier is happening on trigger function, which triggers files parsing DAG:
https://github.com/cern-sis/workflows/blob/elsevier-dataset-parsing/dags/elsevier/trigger_file_processing.py

For all other publishers don't have mappings, so we are extracting everything that is sent.

dags/elsevier/metadata_parser.py
ErnestaP: yes, we are. However, I am using this variable to get a list of files we need to extract from the archive.
In the metadata parser, I am parsing dataset.xml files, which are mappings of articles. These files have a list of records present in the archive, and described paths to XMLs, PDFs, and sometimes, PDF/As. We want just these specific XMLs extracted and passed to further parsing. 
Elsevier sends us files that we don't need to process like svgs, jpegs and etc. In order to make the workflow run more efficient, I am parsing only needed files

dags/elsevier/trigger_file_processing.py
ErnestaP: You are right! Looking closer to it, I noticed we need 2 loops:
1. First loop parses the mapping files (dataset.xml) in order to get paths of xml files and journal info. 
https://github.com/cern-sis/workflows/blob/fb37938761c96d96f61d97739b5f073dcf128503/dags/elsevier/metadata_parser.py#L18-L51
In one dataset.xml file, there are more than one record mapping
2. Second loop iterates through the record files and parses them

dags/common/pull_ftp.py
drjova: I see, but `migrate_files_elsevier` and `migrate_files` are very similar, why don't just push the change to this function instead of propagate it in others?

dags/common/utils.py
```diff
@@ -228,3 +227,63 @@ def process_archive(file_bytes, file_name):
         return process_zip_file(file_bytes, file_name)
     if tarfile.is_tarfile(file_bytes):
         return process_tar_file(file_bytes, file_name)
+
+
+def process_zip_file_elsevier(file_bytes, file_name):
+    file_bytes.seek(0)
+    with zipfile.ZipFile(file_bytes) as zip:
+        for filename in zip.namelist():
+            if "dataset.xml" not in filename:
```
drjova: this is the only difference between `process_zip_file_elsevier` and ` process_zip_file` right? why not just add an `ignore_file_name: list` param to the `process_zip_file` instead of code duplication?

dags/common/pull_ftp.py
ErnestaP: migrate_files impacts ALL the rest of the publishers, so in the end we have to make more changes than should be for the scope of this PR. We can create the task and move it the changes for migrate_files in the other PR. What do you think?

dags/common/utils.py
ErnestaP: True! Good idea, thanks!

dags/common/pull_ftp.py
ErnestaP: the difference between  migrate_files_elsevier and migrate_files is that: migrate_files_elsevier does not extract files, while migrate_files does.
So maybe we can pass the flag extract=False for Elsevier, and then add functionality of migrate_files_elsevier to migrate_files?
for example:
```
def migrate_files(
    archives_names,
    s_ftp: SFTP_FTP_TYPE,
    repo: IRepository,
    logger: PrintLogger,
    extract: bool = False,
):
    logger.msg("Processing files.", filenames=archives_names)
    extracted_or_downloaded_filenames = []
    for archive_name in archives_names:
        logger.msg("Getting file from SFTP.", file=archive_name)
        file_bytes = s_ftp.get_file(archive_name)

        if (
            ".zip" in archive_name
            and zipfile.is_zipfile(file_bytes)
            or ".tar" in archive_name
            and tarfile.is_tarfile(file_bytes)
        ):
            if extract:
                for (archive_file_content, s3_filename) in process_archive(
                    file_bytes=file_bytes, file_name=archive_name
                ):
                    repo.save(s3_filename, io.BytesIO(archive_file_content))
                    if repo.is_meta(s3_filename):
                        extracted_or_downloaded_filenames.append("extracted/" + s3_filename)
                repo.save(archive_name, file_bytes)
            else:
                extracted_or_downloaded_filenames.append("raw" + archive_name)
                repo.save(archive_name, file_bytes)

        else:
            logger.info(
                "File is not zip or tar, processing the next one",
                file_name=archive_name,
            )
            continue
    return extracted_or_downloaded_filenames

```

drjova: we can rename instead of `extract` to `process_archives` and by default will be `True`

dags/common/pull_ftp.py
```diff
@@ -79,22 +92,33 @@ def reprocess_files(repo: IRepository, logger: PrintLogger, **kwargs):
...
+    logger: PrintLogger,
+    publisher: str,
+    **kwargs,
+):
     logger.msg("Force Pulling from SFTP.")
     excluded_directories = kwargs["params"]["excluded_directories"]
     filenames = s_ftp.list_files(excluded_directories=excluded_directories)
+    if publisher == "elsevier":
+        return migrate_files(filenames, s_ftp, repo, logger, process_archives=False)
     return migrate_files(filenames, s_ftp, repo, logger)
```
drjova:
```suggestion
    process_archives = publisher == "elsevier"
    return migrate_files(filenames, s_ftp, repo, logger, process_archives=process_archives)
```

dags/common/pull_ftp.py
```diff
@@ -79,22 +92,33 @@ def reprocess_files(repo: IRepository, logger: PrintLogger, **kwargs):
...
     logger: PrintLogger,
+    publisher: str,
     **kwargs,
 ):
     filenames_pull_params = kwargs["params"]["filenames_pull"]
     filenames = filenames_pull_params["filenames"]
     logger.msg("Pulling specified filenames from SFTP")
+    if publisher == "elsevier":
+        return migrate_files(filenames, s_ftp, repo, logger, process_archives=False)
     return migrate_files(filenames, s_ftp, repo, logger)
```
drjova: same as above

dags/common/pull_ftp.py
```diff
@@ -113,13 +137,19 @@ def _find_files_in_zip(filenames, repo: IRepository):
...
+    **kwargs,
 ):
     logger.msg("Pulling missing files only.")
     excluded_directories = kwargs["params"]["excluded_directories"]
     sftp_files = s_ftp.list_files(excluded_directories=excluded_directories)
     s3_files = repo.get_all_raw_filenames()
     diff_files = list(filter(lambda x: x not in s3_files, sftp_files))
+    if publisher == "elsevier":
+        return migrate_files(diff_files, s_ftp, repo, logger, process_archives=False)
     return migrate_files(diff_files, s_ftp, repo, logger)
```
drjova: same as above

dags/common/utils.py
```diff
@@ -228,3 +227,63 @@ def process_archive(file_bytes, file_name):
         return process_zip_file(file_bytes, file_name)
     if tarfile.is_tarfile(file_bytes):
         return process_tar_file(file_bytes, file_name)
+
+
+def process_zip_file_elsevier(file_bytes, file_name):
```
drjova: we don't need this function anymore

dags/common/utils.py
```diff
@@ -228,3 +227,63 @@ def process_archive(file_bytes, file_name):
...
+        for filename in zip.namelist():
+            if "dataset.xml" not in filename:
+                continue
+            file_prefix = ".".join(file_name.split(".")[:-1])
+            zip_file_content = zip.read(filename)
+            s3_filename = os.path.join(file_prefix, filename)
+            return (zip_file_content, s3_filename)
+
+
+def process_tar_file_elsevier(file_bytes, file_name):
```
drjova: we don't need this function anymore

dags/common/utils.py
```diff
@@ -228,3 +227,63 @@ def process_archive(file_bytes, file_name):
...
+        for filename in tar.getnames():
+            if "dataset.xml" not in filename:
+                continue
+            file_prefix = ".".join(file_name.split(".")[:-1])
+            tar_file_content = tar.extractfile(filename).read()
+            s3_filename = os.path.join(file_prefix, filename)
+            return (tar_file_content, s3_filename)
+
+
+def process_archive_elsevier(file_bytes, file_name):
```
drjova: we don't need this function anymore

dags/common/utils.py
```diff
@@ -228,3 +227,63 @@ def process_archive(file_bytes, file_name):
...
+    files = {}
+    with tarfile.open(fileobj=file_bytes, mode="r") as tar:
+        for filename in tar.getnames():
+            if filename in specific_file:
+                tar_file_content = tar.extractfile(filename).read()
+                files[filename] = tar_file_content
+    return files
+
+
+def process_archive_specific_files(file_bytes, file_name, specific_file):
```
drjova: all these functions are just duplicates with an if statement, we don't need to do that, we can easily add a parameter to the original functions to process just the files we want.

ErnestaP: this one is different, it returns original file name, instead of the s3 key,

dags/elsevier/trigger_file_processing.py
```diff
@@ -0,0 +1,62 @@
...
+        parsed_articles = parser.parse(dataset_file)
+        for parsed_article in parsed_articles:
+            (file_content, full_file_path) = next(
+                process_archive(
+                    file_bytes=file_bytes,
+                    file_name=s3_filename,
+                    only_specific_file=parsed_article["files"]["xml"],
+                    form_s3_key=False,
+                )
+            )
```
drjova: We already have the the file_name right? We can simplify all this by just ignoring the `full_file_path` which is actually the `filename` for this case according to the code of `process_zip_file`
```suggestion
            file_content, _ = next(
                process_archive(
                    file_bytes=file_bytes,
                    file_name=s3_filename,
                    only_specific_file=parsed_article["files"]["xml"],
                    form_s3_key=False,
                )
            )
```

dags/common/utils.py
```diff
@@ -203,28 +202,36 @@ def check_dagrun_state(dagrun: DagRun, not_allowed_states=[], allowed_states=[])
...
     file_bytes.seek(0)
     with tarfile.open(fileobj=file_bytes, mode="r") as tar:
         for filename in tar.getnames():
-            file_prefix = ".".join(file_name.split(".")[:-1])
+            if only_specific_file and only_specific_file not in filename:
+                continue
             tar_file_content = tar.extractfile(filename).read()
+            if not form_s3_key:
+                yield (tar_file_content, filename)
+            file_prefix = ".".join(file_name.split(".")[:-1])
```
drjova: we can do simplify both functions `process_zip_file` and ` process_tar_file` by just passing, `parse_only_filenames=[]` and have:

```python
if parse_only_filenames and filename not in parse_only_filenames
```

This way, will help us add more such cases in the future.

And according to the comment for ` trigger_file_processing_elsevier` we can leave as it is without the `form_s3_key` (there is a typo also 😄 )

This applies to both functions.

dags/elsevier/trigger_file_processing.py
ErnestaP: No, we don't have full_file_path. filename is the name of zip/tar file, and full_file_path is the path of file, which is inside zip/tar

dags/common/utils.py
ErnestaP: I did leave parse_only_filenames=[] on purpose  not list. In this case we don't need an additional loop inside the method, because there is a case when we want to process ONLY ONE file, which is dataset file.
Also passing the kist it will force to change the logic in in trigger_file_processing_elsevier:
I am iterating through names of all the files saved in s3
https://github.com/cern-sis/workflows/blob/3a654b2cbad3274c56fbdad25852925d06ffa342/dags/elsevier/trigger_file_processing.py#L20

Then parsing the mapping (datase.xml) file
https://github.com/cern-sis/workflows/blob/3a654b2cbad3274c56fbdad25852925d06ffa342/dags/elsevier/trigger_file_processing.py#L31


I cannot leave without  form_s3_key, because it is used for saving a file in s3:
https://github.com/cern-sis/workflows/blob/3a654b2cbad3274c56fbdad25852925d06ffa342/dags/common/pull_ftp.py#L38-L42

And just then unzipping the actual  record files:
https://github.com/cern-sis/workflows/blob/3a654b2cbad3274c56fbdad25852925d06ffa342/dags/elsevier/trigger_file_processing.py#L33-L43

Passing an array will make the code harder to read and understand

dags/common/utils.py
```diff
@@ -203,28 +202,32 @@ def check_dagrun_state(dagrun: DagRun, not_allowed_states=[], allowed_states=[])
...
+            if only_specific_file and only_specific_file not in filename:
+                continue
             zip_file_content = zip.read(filename)
+            file_prefix = ".".join(file_name.split(".")[:-1])
             s3_filename = os.path.join(file_prefix, filename)
             yield (zip_file_content, s3_filename)
 
 
-def process_tar_file(file_bytes, file_name):
+def process_tar_file(file_bytes, file_name, only_specific_file):
```
drjova: I would make it optional `only_specific_file` :)

tests/integration/iop/test_dag_pull_ftp.py
```diff
@@ -109,53 +110,55 @@ def test_dag_migrate_from_FTP(iop_empty_repo):
...
             {
-                "pdf": "extracted/2022-09-24T03_01_43_content/1674-1137/1674-1137_46/1674-1137_46_10/1674-1137_46_10_103108/cpc_46_10_103108.pdf",
-                "xml": "extracted/2022-09-24T03_01_43_content/1674-1137/1674-1137_46/1674-1137_46_10/1674-1137_46_10_103108/cpc_46_10_103108.xml",
+                r"pdf": r"extracted/2022-09-24T03_01_43_content/1674-1137/1674-1137_46/1674-1137_46_10/1674-1137_46_10_103108/cpc_46_10_103108.pdf",
+                r"xml": r"extracted/2022-09-24T03_01_43_content/1674-1137/1674-1137_46/1674-1137_46_10/1674-1137_46_10_103108/cpc_46_10_103108.xml",
             },
-            {"xml": "extracted/aca95c.xml/aca95c.xml"},
+            {r"xml": r"extracted/aca95c.xml/aca95c.xml"},
         ]
+        assert sorted(iop_empty_repo.find_all(), key=lambda x: x.get("xml", "")) == sorted(expected_files, key=lambda x: x.get("xml", ""))
```
ErnestaP: For some reason this test started to fail because expected_files are read with new lines:
```
AssertionError: assert [{'pdf': 'extracted/2022-07-30T03_02_01_content/1674-1137/1674-1137_46/1674-1137_46_8/1674-1137_46_8_085001/cpc_46_8_085001.pdf',\n  'xml'
```

dags/common/utils.py
drjova: @ErnestaP ☝️

dags/elsevier/elsevier_file_processing.py
```diff
@@ -6,27 +6,23 @@
...
-    file_name = kwargs["params"]["file_name"]
     xml_bytes = base64.b64decode(encoded_xml)
     xml = parse_without_names_spaces(xml_bytes.decode("utf-8"))
     parser = ElsevierParser()
     parsed = parser.parse(xml)
-    parsed["source_file_path"] = file_name
-    return parsed
+    if "params" not in kwargs and "metadata" not in kwargs["params"]:
+        raise Exception("Record is missing metadata. Exiting run.")
+    metadata = kwargs["params"]["metadata"]
```
drjova: a more python way:

```
try:
      kwargs["params"]["metadata"]
except KeyError:
  
```

dags/elsevier/metadata_parser.py
```diff
@@ -101,17 +115,20 @@ def _get_license(self, article):
...
-        }
-        xml = {
-            "filetype": "xml",
-            "path": os.path.join(
+        if "A.tar" in self.file_path:
+            self.file_path = self.file_path.split("A.")[:-1][0]
+        if ".zip" in self.file_path:
+            self.file_path = self.file_path.split(".zip")[:-1][0]
+        if "raw" in self.file_path:
+            self.file_path = os.path.join(*self.file_path.split("/")[1:])
```
drjova: why not something like `os.path.dirname(os.path.realpath())` or even better take advantage of python 3?

dags/common/utils.py
ErnestaP: but it's optional already in the main function, from where the process_tar and zip is called:
https://github.com/cern-sis/workflows/blob/33024a61c390aeb8ec2c6610d073e5190c7b45e6/dags/common/utils.py#L229

Process_tar/zip is never called separately, its always called from process_archive
Should I make it still optional in these functions, in our case?

drjova: yep otherwise, if we have to use the functions directly, we will always have to pass the specific file :)

drjova: and actually we could just:

```

def process_archive(file_bytes, file_name, **kwargs):
```

and pass them down to the other functions for simplicity

dags/common/utils.py
```diff
@@ -203,28 +202,34 @@ def check_dagrun_state(dagrun: DagRun, not_allowed_states=[], allowed_states=[])
...
             zip_file_content = zip.read(filename)
+            file_prefix = ".".join(file_name.split(".")[:-1])
             s3_filename = os.path.join(file_prefix, filename)
             yield (zip_file_content, s3_filename)
 
 
-def process_tar_file(file_bytes, file_name):
+def process_tar_file(file_bytes, file_name, **kwargs):
     file_bytes.seek(0)
+    only_specific_file = kwargs.get("only_specific_file", None)
```
drjova: no need of `None` by default it returns `None` :)

dags/elsevier/elsevier_file_processing.py
```diff
@@ -6,27 +6,25 @@
...
     xml = parse_without_names_spaces(xml_bytes.decode("utf-8"))
     parser = ElsevierParser()
     parsed = parser.parse(xml)
-    parsed["source_file_path"] = file_name
-    return parsed
+    try:
+        kwargs["params"]["metadata"]
+    except KeyError:
+        raise Exception("Record is missing metadata. Exiting run.")
+    metadata = kwargs["params"]["metadata"]
```
drjova:
```suggestion
    try:
        metadata = kwargs["params"]["metadata"]
    except KeyError:
        raise Exception("Record is missing metadata. Exiting run.")
```

dags/elsevier/elsevier_file_processing.py
```diff
@@ -6,27 +6,25 @@
...
 from jsonschema import validate
 
 
 def parse_elsevier(**kwargs):
-    if "params" not in kwargs or "file" not in kwargs["params"]:
-        raise KeyError("There was no 'file' parameter. Exiting run.")
-    if "params" not in kwargs or "file_name" not in kwargs["params"]:
-        raise KeyError("There was 'file_name' parameter. Exiting run.")
+    if "params" not in kwargs and "file" not in kwargs["params"]:
+        raise Exception("There was no 'file' parameter. Exiting run.")
```
drjova: but that's different from the above, why removing the `file_name`?

ErnestaP: File is XML file content, we can rename it, it comes from a [trigger_file_processing_elsevier](https://github.com/cern-sis/workflows/blob/33024a61c390aeb8ec2c6610d073e5190c7b45e6/dags/elsevier/trigger_file_processing.py#L55).
[trigger_file_processing](https://github.com/cern-sis/workflows/blob/33024a61c390aeb8ec2c6610d073e5190c7b45e6/dags/common/pull_ftp.py#L182) also sends file content in variables named file. I will create an issue to update it, since it touches all the rest of publishers 
However, I did not catch the comment "why removing the file_name" ?

drjova (approved):

PR #180, opened 2023-11-07
Dockerfile: added airflow constrains

ErnestaP:


requirements-airflow.txt
```diff
@@ -1,2 +1,3 @@
-psycopg2==2.9.6
+psycopg2-binary==2.9.6
+connexion==2.14.2
```
pamfilos: not needed. should be taken care of by the constraints..if it is not installed by the constraints, we should investigate which library needs it, and check if we should install differently

ErnestaP: Constraints files are requirements files that only control which version of a requirement is installed, not whether it is installed or not. Their syntax and contents is nearly identical to Requirements Files.
https://pip.pypa.io/en/stable/user_guide/#constraints-files

pamfilos: ok ..what i mean is:

for postgres, we shouldnt have `psycopg2==2.9.6` or `psycopg2-binary==2.9.6`, but install it as `apache-airflow[celery, postgres]==2.6.0`

this way it will install all dependencies that are needed for postgres, and use the constraints for choosing a version

pamfilos: in the same way, we should try to see which package tries to install `connexion` and why the constraints dont work

pamfilos: in the same way, we should try to see which package tries to install `connexion` and why the constraints do not work

pamfilos (approved):

PR #181, opened 2023-11-27
Logs cleaning

ErnestaP:


dags/common/pull_ftp.py
```diff
@@ -186,4 +186,4 @@ def trigger_file_processing(
 
 
 def _generate_id(publisher: str):
-    return datetime.utcnow().strftime(f"{publisher}_%Y-%m-%dT%H:%M:%S.%f")
+    return datetime.utcnow().strftime(f"{publisher}__%Y-%m-%dT%H:%M:%S.%f")
```
drjova: why this change? it will affect all the ids from now on?

ErnestaP: because we have triggering the pull_ftp/fecth_api dags' tasks, we have ids generated with double dashes and then date, for example:
scheduled__2023-11-27T10:00:00+00:00
manual__2023-11-27T10:28:35.902469+00:00

Since the process dags' tasks we are triggering with a trigger, we are specifying the ids manually with _generate_id id function. To keep the names in consistent pattern allows me to select the run from DB and extract dates correctly:
https://github.com/cern-sis/workflows/blob/990a8a2a3ee69688d85281c95d4de5c6829ee1ef/dags/clean/utils.py#L15

dags/clean/utils.py
```diff
@@ -0,0 +1,48 @@
...
+    delete_query = f"""
+        DELETE FROM task_instance
+        WHERE run_id='{run_id}'
+        """
+    src_pg = PostgresHook(postgres_conn_id="airflow_db")
+    conn = src_pg.get_conn()
+    with conn.cursor() as cursor:
+        cursor.execute(delete_query)
+        conn.commit()
+        print("Rows deleted successfully!")
```
drjova: we can just log

dags/common/pull_ftp.py
drjova: @ErnestaP thanks, but still I cannot understand. Why instead of changing the id, we change just the query?

ErnestaP: In order to get results we should have a template, for how the run_ids are named. To have it consistent with autogenerated ids (airflow run_ids) and our named run_ids (runs that we are triggering by the trigger tasks:
https://github.com/cern-sis/workflows/blob/990a8a2a3ee69688d85281c95d4de5c6829ee1ef/dags/common/pull_ftp.py#L159-L185
)
the template is to have a double underscore before the date

ErnestaP: I don't see the point make the query more complex than it should be. It is easier, cleaner and better approach to make the run_ids consistet

ErnestaP: and I see I did not replied  to your question  directly:
_it will affect all the ids from now on?_
Yes, all the tasks will have double underscore  before the date

drjova: I still don't understand apologies. We already have a template with 1 underscore, right? Why changing? Do we set somewhere in airflow to add 2 underscores? From where these "autogenerated ids" are set?

ErnestaP: Can we have a call? will be easier to explain

ErnestaP: No, we don't set the double underscore anywhere, it comes from Airflow itself.
Current implementation is wrong and not consistent, in my opinion, and I will try to show you why.

Let's say if the dag is triggered not by our written trigger but by schedular, the name of the run_id is:
scheduled__...
It comes by default from airflow, please the a look into  documentation : https://airflow.apache.org/docs/apache-airflow/1.10.14/_modules/airflow/models/dagrun.html#DagRun.run_id

I want to make our trigger created run_ids consistent with airflow run_ids. So, I think that adding one more underscore is a great and easy solution:
1) the name is consistent with Airflow naming
2) The SQL query for constructing logs names is simple. SQL query has to extract date from run_id, but having single underscore, it is harder, since we have two different variations:
Default ids, coming from airflow
a) scheduled__date, manual__date
and your custom ones
b) iop_date, oup_date, springer_date

While when we keeping it consistent with airlfow ids, we can have just one variation:
a)  scheduled__date,  manual__date, iop__date, oup__date, springer__date

In my personal opinion, the run_ids value in our custom trigger were not fully analyzed, at the time it was written

requirements-airflow.txt
```diff
@@ -1,2 +1,3 @@
 -c https://raw.githubusercontent.com/apache/airflow/constraints-2.6.0/constraints-3.10.txt
 apache-airflow[celery, postgres]==2.6.0
+apache-airflow-providers-postgres==5.4.0
```
drjova: we don't need this anymore

dags/clean/clean.py
```diff
@@ -0,0 +1,47 @@
...
+    ' _ {{}} "$current_date" \;
+"""
+
+cleanup_task = BashOperator(
+    task_id='cleanup_logs',
+    bash_command=cleanup_command,
+    dag=dag,
+)
+
+cleanup_task
```
drjova: Why don't we keep the same structure as the other dags? This is a different way to define it :)

tests/units/clean/test_clean.py
```diff
@@ -0,0 +1,39 @@
...
+    logs_date = datetime.datetime.now()
+    log_path = f'tests/units/clean/logs/dag_id=test/run_id=test__{logs_date.strftime("%Y-%m-%dT%H:%M:%S.%f%z")}/task_id=test_task/attempt=1.log'
+    os.makedirs(os.path.dirname(log_path), exist_ok=True)
+    try:
+        with open(log_path, "w") as f:
+            f.write("test log")
+        return True
+    except:
+        logging.error('Failed file creation for tests')
+        pass
```
drjova: I would recommend not to  write on the file system in the tests, you could use virtual temp files in python to test the same functionality :)

tests/units/clean/test_clean.py
```diff
@@ -0,0 +1,39 @@
...
+    assert dag is not None
+    assert len(dag.tasks) == 1
+
+
+@freeze_time("2023-09-20")
+@fixture
+def old_temp_dir(tmpdir, tmp_path):
+    logs_date = datetime.utcnow().astimezone(timezone.utc)
+    log_path = tmpdir.join(
+        f'logs/dag_id=test/run_id=test__{logs_date.strftime("%Y-%m-%dT%H:%M:%S.%f%z")}/task_id=test_task/attempt=1.log'
```
drjova, line 28: here your example has `z`. Also we are testing against different `_generate_id` would it work?

ErnestaP: Previous versions of the script checked the folder name, the current bash script does not care about the folder name and the format of the date in it. 
It filters files by the modification date, so we don't care about date format in folder name, when we delete files

ErnestaP: here your example has z - I made it consistent with airflow generated run_id=.. value

drjova (approved):

PR #183, opened 2023-12-01
Task create_or_update: run after record validation

ErnestaP:


dags/oup/oup_process_file.py
```diff
@@ -52,7 +53,7 @@ def enrich_file(enhanced_file):
 
     @task()
     def validate_record(enriched_file):
-        oup_validate_record(enriched_file)
+        return enriched_file and oup_validate_record(enriched_file)
```
drjova: this will return a bool which is different from what we need for `create_or_update`

ErnestaP: no, it will return the record:
https://github.com/cern-sis/workflows/blob/46158b9ec68c8ab5a1723154408235b4ecc2dc32/dags/oup/oup_process_file.py#L34-L37

drjova: not if enriched_file is false

ErnestaP: if enriched_files is false it will fail on validation and raise an error, so it will not even reach the create_and_update task

drjova: then why  do we need to check for the `enriched_file`?

ErnestaP: for example, to be sure that enriched_file is not empty variable

drjova: Yes which is the same with `return enriched_file and oup_validate_record(enriched_file)`

ErnestaP: yes, that's the check you were referring to, no? 
```
then why do we need to check for the enriched_file?
```

ErnestaP: I don't understand what is wrong here. Can you please give me more details, what the problem is pointed here?

drjova (approved):

PR #184, opened 2023-12-04
APS: bug fix

ErnestaP:


tests/units/aps/test_utils.py
```diff
@@ -57,10 +58,10 @@ def test_save_file_in_s3():
     assert key == expected_key
 
 
-def test_split_json():
-    doi = "10.1103/PhysRevLett.126.153601"
-    today = datetime.now().strftime("%Y-%m-%dT%H:%M")
+@freeze_time("2023-12-04")
+def test_split_json(monkeypatch):
+    monkeypatch.setenv("TZ", "GMT")
```
drjova: Unfortunately, this will not work either since if we change, the server's timezone might be different. We don't need this https://github.com/cern-sis/workflows/commit/d2f9e84841cd9ae01ce0485a968a13c47d44a7b5#diff-70b5a470d6615a3d2f378c02dd505cac0aca0f865f188a88f95d0a69912bd630R189 in the first place. If we remove it, it will work with just mocking with freeze time.

ErnestaP: But then how I need to get date in the format of something like this: 2023-12-04T10:01:21+00:00?

ErnestaP: I wanted to be consistent with run_id namings, like run_id=manual__2023-12-04T10:01:21+00:00

drjova: removing the https://github.com/cern-sis/workflows/commit/d2f9e84841cd9ae01ce0485a968a13c47d44a7b5#diff-70b5a470d6615a3d2f378c02dd505cac0aca0f865f188a88f95d0a69912bd630R189 and just add the following should work

```
@freeze_time("2023-12-04 10:00")
def test_split_json(monkeypatch):
    ids_and_articles = split_json(repo=MockedRepo(), key="key/key")
    expected_id = f"APS__2023-12-04T10:00:00.000000"
    assert ids_and_articles[0]["id"] == expected_id
    assert len(ids_and_articles) == 1
```

ErnestaP: I will repeat myself, but I need a format with  +. Not 2023-12-04T10:00:00.000000, but with 2023-12-04T10:00:00+00:00

ErnestaP: But if it takes too much effort or is not even possible, I can keep without it. Just I prefer to have consistent run_id naming

drjova: if you really need it:

```
def _generate_id(publisher: str):
    logs_date = datetime.now(timezone.utc)
    return f'{publisher}__{logs_date.strftime("%Y-%m-%dT%H:%M:%S.%f%z")}'

```

```
@freeze_time("2023-12-04 10:00")
def test_split_json():
    ids_and_articles = split_json(repo=MockedRepo(), key="key/key")
    expected_id = f"APS__2023-12-04T10:00:00.000000+0000"
    assert ids_and_articles[0]["id"] == expected_id
    assert len(ids_and_articles) == 1
```

drjova (approved):

PR #185, opened 2023-12-07
Elsevier: parse a group affiliation if there is no affid assigned for…

ErnestaP:


drjova:
@ErnestaP instead of changing the code, wouldn't it be enough to filter out the `{value: None}`?

ErnestaP:
Just noting what I said this morning: we cannot because the authors should have affiliation parsed differently if there is no affid associated with them. We need to take the whole affiliation group. Leaving the code the same and just clean None values, will leave authors without affiliations. While the affiliations are actually in the metadata, just have to be parsed in another way

tests/units/elsevier/test_elsevier_parser.py
```diff
@@ -11553,17 +11580,19 @@ def parsed_articles(parser, articles):
...
         param(
-            ["137730", "138173", "137649"],
+            ["137730", "138173", "137649", "138109"],
             "journal_artid",
             id="test_journal_artid",
         ),
     ],
 )
 def test_elsevier_parsing(parsed_articles, expected, key):
+    print(parsed_articles[3]["authors"])
```
drjova: we don't need this :)

PR #186, opened 2023-12-11
Elsevier: publication_info added artid, journal doctype, volume

ErnestaP:


dags/common/utils.py
```diff
@@ -72,7 +72,7 @@ def parse_to_int(value):
         logger.error("Cannot parse to integer", value=value)
 
 
-def extract_text(article, path, field_name, dois):
+def extract_text(article, path, field_name, dois="No DOI passed"):
```
drjova: I think this should be part of the doi extraction, we don't need it for the log

drjova (approved):

PR #193, opened 2024-01-26
common: create_or_update_article update fix

ErnestaP:


dags/common/utils.py
```diff
@@ -248,14 +248,37 @@ def create_or_update_article(data):
...
+                return update_response.json()
+            except requests.HTTPError:
+                logger.error(f"Update failed: {update_response.content}")
+                raise requests.HTTPError(
+                    f"Update failed: {update_response.content}",
+                    response=update_response,
+                )
+        raise requests.HTTPError(
+            f"Create failed: {error_content.get('message')}", response=e.response
+        )
```
drjova: nice, I would suggest to split the logic of create and update in 2 different functions

dags/common/utils.py
```diff
@@ -248,14 +248,41 @@ def create_or_update_article(data):
...
+        logger.error(f"Error message: {error_content.get('message')}")
+        if "article_id" in error_content:
+            return update_article(data, backend_url, error_content, headers)
+        raise requests.HTTPError(
+            f"Create failed: {error_content.get('message')}", response=e.response
+        )
+
+
+def update_article(data, backend_url, error_content, headers):
+    article_id = error_content["article_id"]
```
drjova: 1. since we have a check for `article_id` https://github.com/cern-sis/workflows/pull/193/files#diff-caa66940a5e0989c3d3a1a153a6658ead25d65b7f12ccef0d0efb3d46c032cecR264, there is no need to pass the `error_content` around.  We can just pass the `article_id`
2. we can easily reconstruct the backend_url from the env vars
3. also the `headers`, we can create a function to use in both of them or we can just reconstruct them inside the function, no need to pass it around.

At the end the function could have only 2 params:

```python
def update_article(article_id, data):
....
```

ErnestaP:
No need this implementation, because the update happens in POST (it means **create**) method

PR #207, opened 2024-06-04
Airflow upgrade

ErnestaP:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM apache/airflow:2.6.0-python3.10
+FROM apache/airflow:2.8.3-python3.8
```
jimil749, line 1: can we use this image - `registry.cern.ch/cern-sis/airflow-base:2.8.3`? This is the base airflow image to be used across all the dags. It is airflow - 2.8.3 and python 3.10

pamfilos (commented):
lets remove the `scoap3_` prefix now that we have separate clusters

ErnestaP:
Closing this PR. These changes are not needed at all, since python version is now 3.10

PR #211, opened 2024-07-09
Orcids parsing

ErnestaP:


dags/oup/parser.py
```diff
@@ -183,20 +184,19 @@ def _get_authors(self, article):
...
-                        "affiliations": full_affiliation,
-                    }
-                )
+                author = {
+                    "surname": surname,
+                    "given_names": given_names,
+                    "email": email,
+                    "affiliations": full_affiliation,
+                }
+                orcid and author.update({"orcid": orcid})
```
pamfilos: Lets change to be with `if..`. its cleaner

tests/units/elsevier/test_elsevier_parser.py
```diff
@@ -81,11 +85,11 @@ def enhanced_articles(parser, parsed_articles):
                         ],
                     },
                     {
-                        "surname": "Adamová",
+                        "surname": "Adamov\u00e1",
```
pamfilos: why did we need to change these values?

pamfilos (changes requested):

pamfilos (approved):

PR #238, opened 2024-12-06
feat(APS): Author parsing through xml files

Lorenzovagliano:
Issue #376 (https://github.com/cern-sis/issues-scoap3/issues/376)

Issue #383 (https://github.com/cern-sis/issues-scoap3/issues/383)


pamfilos:
it actually seems that the ROR id is included in the value of the affiliations. Maybe we can try and extract them from there, also removing the `<a>` tags also from the value

pamfilos:
closing as superseded by #277

PR #252, opened 2025-04-08
fix(Utils): Parse XMLs with wrong namespaces

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-scoap3/issues/433


tests/units/common/test_common_utils.py
```diff
@@ -1,9 +1,37 @@
...
+import sys
 from unittest import mock
 
 import pytest
 import requests
-from common.utils import create_or_update_article
+from common.utils import create_or_update_article, remove_xml_namespaces
+
+
+class FunctionCallTracker:
```
pamfilos: remove please , since not used

Lorenzovagliano: Ahh, forgot about that. Yes.

pamfilos (approved):

PR #265, opened 2025-06-26
feat(IOP): Use SSH key for sftp auth

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-scoap3/issues/458

This should be thoroughly tested on QA before being pushed to prod.

We must also send our public SSH key to IOP.


dags/iop/sftp_service.py
```diff
@@ -8,7 +11,35 @@ def __init__(self):
...
             host=os.getenv("IOP_FTP_HOST", "localhost"),
             username=os.getenv("IOP_FTP_USERNAME", "airflow"),
-            password=os.getenv("IOP_FTP_PASSWORD", "airflow"),
+            password=None,
             port=int(os.getenv("IOP_FTP_PORT", "2222")),
             dir=os.getenv("IOP_FTP_DIR", "upload/iop"),
         )
+        self.private_key_content = os.getenv("IOP_FTP_SSH_PRIVATE_KEY")
+
+    def _SFTPService__connect(self):
```
drjova: I would move the logic in the SFTPService to be reusable in other publishers if needed. If it has an SSH key, connect with it; otherwise, connect with a password. WDYT?

tests/conftest.py
```diff
@@ -1,8 +1,39 @@
+import os
 import unittest
+from unittest import mock
 
 import airflow
 import pytest
 
+FAKE_KEY = """-----BEGIN OPENSSH PRIVATE KEY-----
```
drjova: could we export in the env of the tests instead of keeping two copies?

drjova (changes requested):

dags/common/sftp_service.py
```diff
@@ -16,25 +17,46 @@ def __init__(
...
-        client.connect(
-            banner_timeout=200,
-            hostname=self.host,
-            username=self.username,
-            password=self.password,
-            port=self.port,
-        )
+
+        if self.private_key_content:
+            if not self.private_key_content:
```
drjova: it will never go here 🤔 it will go to `else` otherwise

dags/common/sftp_service.py
```diff
@@ -16,25 +17,46 @@ def __init__(
...
+                port=self.port,
+            )
+        else:
+            client.connect(
+                banner_timeout=200,
+                hostname=self.host,
+                username=self.username,
+                password=self.password,
+                port=self.port,
+            )
```
drjova, line 50: we can simplify the logic here, WDYT? 


```python
connect_kwargs = {
    "banner_timeout": 200,
    [.....]
}


# and then something like


if self.private_key_content:
    key_file = io.StringIO(self.private_key_content)
    connect_kwargs["pkey"] = paramiko.RSAKey.from_private_key(key_file)
elif self.password:
    connect_kwargs["password"] = self.password
else:
    raise ValueError("No keys ..........")

client.connect(**connect_kwargs)
```

drjova (changes requested):
Thanks @Lorenzovagliano few comments

drjova (approved):
Let's try this on QA, it should affect prod anyway as long as we don't set the key

PR #267, opened 2025-07-02
fix(Build): Remove minimum airflow version

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-scoap3/issues/460

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -253,17 +253,6 @@ services:
         function ver() {
           printf "%04d%04d%04d%04d" $${1//./ }
         }
-        airflow_version=$$(AIRFLOW__LOGGING__LOGGING_LEVEL=INFO && gosu airflow airflow version)
-        airflow_version_comparable=$$(ver $${airflow_version})
-        min_airflow_version=2.2.0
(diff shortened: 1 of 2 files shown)


docker-compose.yaml
```diff
@@ -255,8 +255,6 @@ services:
         }
         airflow_version=$$(AIRFLOW__LOGGING__LOGGING_LEVEL=INFO && gosu airflow airflow version)
         airflow_version_comparable=$$(ver $${airflow_version})
-        min_airflow_version=2.2.0
-        min_airflow_version_comparable=$$(ver $${min_airflow_version})
         if (( airflow_version_comparable < min_airflow_version_comparable )); then
```
drjova: I guess you don't need this check either. But why it's a problem? Are we running airflow<2.2 anywhere?

drjova (changes requested):
A small comment, thanks!

drjova (approved):

drjova:
Let's mark it as xfail

drjova:
@Lorenzovagliano there is a sorting issue

PR #270, opened 2025-07-17
feat(common): Refactor country processing

Lorenzovagliano:
ref: https://github.com/cern-sis/issues-scoap3/issues/464

### DIFF
--- a/dags/common/constants.py
+++ b/dags/common/constants.py
@@ -251,3 +251,55 @@ INSTITUTIONS_AND_COUNTRIES_MAPPING = OrderedDict(
         ("R.O.C", "Taiwan"),
     ]
 )
+
+SPECIAL_CASES = {
+    "european organization for nuclear research": "CERN",
+    "conseil européen pour la recherche nucléaire": "CERN",
+    "stanford linear accelerator center": "United States",
+    "joint institute for nuclear research": "JINR",
+    "fermilab": "United States",
+    "jinr": "JINR",
+    "infn": "Italy",
+    "cern": "CERN",
+    "desy": "Germany",
+    "fnal": "United States",
+    "slack": "United States",
+    "kek": "Japan",
+    "hong kong": "Hong Kong",
+    "hong-kong": "Hong Kong",
+    "republic of china": "Taiwan",
+    "taiwan": "Taiwan",
+    "roc": "Taiwan",
+    "r.o.c": "Taiwan",
+    "people's republic of china": "China",
+    "u.s.a": "United States",
+    "america": "United States",
+    "uk": "United Kingdom",
+    "u.k": "United Kingdom",
+    "northern cyprus": "Türkiye",
+    "north cyprus": "Türkiye",
+    "belgique": "Belgium",
+    "deutschland": "Germany",
+    "democratic people's republic of korea": "North Korea",
+    "dpr korea": "North Korea",
+    "dpr. korea": "North Korea",
+    "d.p.r. korea": "North Korea",
+    "dprk": "North Korea",
+    "d.p.r.k": "North Korea",
+    "korea": "South Korea",
+    "republic of korea": "South Korea",
+    "niger": "Niger",
+    "belgrade": "Serbia",
+    "philipines": "Philippines",
+}
+
+SPECIAL_PATTERNS = re.compile(
+    r"\b("
+    + "|".join(
+        re.escape(k) for k in sorted(SPECIAL_CASES.keys(), key=len, reverse=True)
+    )
+    + r")\b",
+    re.IGNORECASE,
+)
+
+CHAR_REPLACEMENTS = str.maketrans(",-", "  ")
--- a/dags/common/utils.py
+++ b/dags/common/utils.py
@@ -12,22 +12,21 @@ from stat import S_ISDIR, S_ISREG
 
 import backoff
 import country_converter as coco
+import pycountry
 import requests
 from airflow.models.dagrun import DagRun
 from airflow.utils.state import DagRunState
 from common.constants import (
     BY_PATTERN,
     CDATA_PATTERN,
+    CHAR_REPLACEMENTS,
     COUNTRIES_DEFAULT_MAPPING,
     CREATIVE_COMMONS_PATTERN,
-    INSTITUTIONS_AND_COUNTRIES_MAPPING,
     LICENSE_PATTERN,
+    SPECIAL_CASES,
+    SPECIAL_PATTERNS,
 )
-from common.exceptions import (
-    FoundMoreThanOneMatchOrNone,
-    UnknownFileExtension,
-    UnknownLicense,
-)
+from common.exceptions import UnknownFileExtension, UnknownLicense
 from inspire_utils.record import get_value
 from structlog import get_logger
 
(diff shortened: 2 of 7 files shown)


dags/common/utils.py
```diff
@@ -321,37 +320,65 @@ def create_or_update_article(data):
...
 
-def parse_country_from_value(affiliation_value):
-    for key, val in INSTITUTIONS_AND_COUNTRIES_MAPPING.items():
-        if re.search(r"\b%s\b" % key, affiliation_value, flags=re.IGNORECASE):
-            return val
-    country = affiliation_value.split(",")[-1].strip()
-    for key, val in COUNTRIES_DEFAULT_MAPPING.items():
-        if re.search(r"\b%s\b" % key, country, flags=re.IGNORECASE):
-            return val
+def unwrap_coco_result(res):
```
drjova: coco?

dags/common/utils.py
```diff
@@ -321,37 +320,65 @@ def create_or_update_article(data):
...
+    return res
+
+
+def initialize_special_cases():
+    global SPECIAL_PATTERNS
+    sorted_keys = sorted(SPECIAL_CASES.keys(), key=len, reverse=True)
+    SPECIAL_PATTERNS = [
+        (re.compile(rf"\b{re.escape(key)}\b"), SPECIAL_CASES[key])
+        for key in sorted_keys
+    ]
```
drjova: It’s not a good practice to use globals in general. Why not just calculate and return the special patterns? If they are unchanged, we can simply calculate them in a constant file and import it. This way, it will be calculated only once when the module is import it.

dags/common/utils.py
```diff
@@ -321,37 +320,65 @@ def create_or_update_article(data):
...
+        if pattern.search(norm):
+            return country
+
+    raw_iso2 = cc.convert(names=norm, to="ISO2")
+    iso2 = unwrap_coco_result(raw_iso2)
+    if iso2:
+        raw_short = cc.convert(names=iso2, to="name_short")
+        short = unwrap_coco_result(raw_short)
+        if short:
+            return short
```
drjova: Let's separate this to a different function

dags/common/utils.py
```diff
@@ -321,37 +320,65 @@ def create_or_update_article(data):
...
+        (re.compile(rf"\b{re.escape(key)}\b"), SPECIAL_CASES[key])
+        for key in sorted_keys
+    ]
+
+
+def parse_country_from_value(value):
+    if not SPECIAL_PATTERNS:
+        initialize_special_cases()
+
+    norm = value.lower().translate(CHAR_REPLACEMENTS)
```
drjova: `normalized_name`?

dags/common/utils.py
```diff
@@ -321,37 +320,65 @@ def create_or_update_article(data):
...
+            try:
+                matches = pycountry.countries.search_fuzzy(phrase)
+                if matches:
+                    return unwrap_coco_result(
+                        cc.convert(names=matches[0].alpha_2, to="name_short")
+                    )
+            except LookupError:
+                continue
+
+    return None
```
drjova: and this one to a separate function, also I would add more descriptive names, for example `normilized_name_parts` `normilized_name_partslength`, `normilized_name_parts_length`

dags/common/utils.py
```diff
@@ -321,37 +320,65 @@ def create_or_update_article(data):
...
+    iso2 = unwrap_coco_result(raw_iso2)
+    if iso2:
+        raw_short = cc.convert(names=iso2, to="name_short")
+        short = unwrap_coco_result(raw_short)
+        if short:
+            return short
+
+    parts = norm.split()
+    plen = len(parts)
+    max_len = min(4, plen)
```
drjova: why 4? 🤔 It might be too small

drjova (changes requested):
I like the approach, few comments :)

dags/common/utils.py
```diff
@@ -321,37 +319,75 @@ def create_or_update_article(data):
...
-        if re.search(r"\b%s\b" % key, affiliation_value, flags=re.IGNORECASE):
-            return COUNTRIES_DEFAULT_MAPPING[key]
+
+def unwrap_country_converter_result(converter_result):
+    if isinstance(converter_result, (list, tuple)):
+        for item in converter_result:
+            if isinstance(item, str) and item.lower() != "not found":
+                return item
+        return None
+    if not converter_result or str(converter_result).lower() == "not found":
```
```suggestion
    if not converter_result or converter_result.lower() == "not found":
```

should be enough if converter_result cannot be something else

dags/common/utils.py
```diff
@@ -321,37 +319,75 @@ def create_or_update_article(data):
...
+
+def match_with_country_converter_results(normalized_value):
+    raw_iso2_result = cc.convert(names=normalized_value, to="ISO2")
+    iso2_code = unwrap_country_converter_result(raw_iso2_result)
+    if iso2_code:
+        raw_short_name = cc.convert(names=iso2_code, to="name_short")
+        country_short_name = unwrap_country_converter_result(raw_short_name)
+        if country_short_name:
+            return country_short_name
+    return None
```
drjova: exit early :) by default python functions return None, no need to explicitly return


```suggestion
    if not iso2_code:
        return None
    raw_short_name = cc.convert(names=iso2_code, to="name_short")
    country_short_name = unwrap_country_converter_result(raw_short_name)
    if country_short_name:
        return country_short_name
```

dags/common/constants.py
```diff
@@ -251,3 +251,51 @@
...
+    "republic of korea": "South Korea",
+    "niger": "Niger",
+    "belgrade": "Serbia",
+    "philipines": "Philippines",
+}
+
+SPECIAL_PATTERNS = [
+    (re.compile(rf"\b{re.escape(key)}\b"), SPECIAL_CASES[key])
+    for key in sorted(SPECIAL_CASES.keys(), key=len, reverse=True)
+]
```
drjova: we could build just one big pattern instead of x small ones and we can add `re.IGNORECASE` to avoid the need of lowercase
```
"|".join(......)
```

dags/common/utils.py
```diff
@@ -321,37 +319,75 @@ def create_or_update_article(data):
...
 
-def parse_country_from_value(affiliation_value):
-    for key, val in INSTITUTIONS_AND_COUNTRIES_MAPPING.items():
-        if re.search(r"\b%s\b" % key, affiliation_value, flags=re.IGNORECASE):
-            return val
-    country = affiliation_value.split(",")[-1].strip()
-    for key, val in COUNTRIES_DEFAULT_MAPPING.items():
-        if re.search(r"\b%s\b" % key, country, flags=re.IGNORECASE):
-            return val
+def check_special_cases(normalized_value):
```
drjova, line 323: and here we take advantage if the one pattern and do 
```
SPECIAL_PATTERNS.search(text).group(1) ......
```

drjova: please handle the None case etc 🙃

dags/common/utils.py
```diff
@@ -321,37 +319,75 @@ def create_or_update_article(data):
...
+            return country_short_name
+    return None
+
+
+def fuzzy_search_country(normalized_value):
+    normalized_name_parts = normalized_value.split()
+    normalized_name_parts_length = len(normalized_name_parts)
+    max_phrase_length = min(6, normalized_name_parts_length)
+
+    for skip_from_end in range(normalized_name_parts_length):
```
drjova, line 357: I might go for the longest phrases first `range(max_phrase_length, 0, -1):` and then the from start to phrase_length. WDYT?

drjova (changes requested):
few more changes :)

PR #278, opened 2026-03-03
fix: fix _get_text_value for xml parser

PascalEgn:
fixes a parsing issue which leads to some title fields staying empty, see https://github.com/cern-sis/issues-scoap3/issues/483 for examples.

The `.text` attribute used before only returns the text directly inside the element before the first child. This is in general not what we want and also leads to the return value None in cases like:
```
<ArticleTitle>
  <Emphasis>stanhf</Emphasis>: <Emphasis>HistFactory</Emphasis>
  models in the probabilistic programming language
  <Emphasis>Stan</Emphasis>
</ArticleTitle>
```

### DIFF
--- a/dags/common/parsing/xml_extractors.py
+++ b/dags/common/parsing/xml_extractors.py
@@ -19,6 +19,7 @@ class TextExtractor(IExtractor):
         extra_function=lambda s: s,
         prefixes=None,
         all_content_between_tags=False,
+        use_itertext=False,
         remove_tags=False,
     ):
         super().__init__(destination)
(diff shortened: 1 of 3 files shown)


dags/common/parsing/xml_extractors.py
```diff
@@ -34,7 +34,7 @@ def __init__(
 
     def _get_text_value(self, raw_value):
         try:
-            return raw_value.text
+            return "".join(raw_value.itertext()).strip()
```
pamfilos: lets create a new `_get_itertext_value` method for the new parse logic

pamfilos (approved):

PR #303, opened 2026-05-21
workflows(parser): Clean up APS affiliation organization text

namollayo:
* Remove address and country data from the organization field

* ref: #524


PascalEgn:
@namollayo please squash, then we can merge 👍

PR #304, opened 2026-07-02
s3: Fix uploaded file content-type headers

namollayo:
* Apply accurate content-type depending on file extension
* ref: cern-sis/issues-scoap3#532

### DIFF
--- a/dags/common/scoap3_s3.py
+++ b/dags/common/scoap3_s3.py
@@ -1,4 +1,5 @@
 import logging
+import mimetypes
 import os
 from io import BytesIO
 from uuid import uuid4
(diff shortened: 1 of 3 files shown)


dags/common/scoap3_s3.py
```diff
@@ -43,13 +60,15 @@ def copy_file(self, source_bucket, source_key, prefix=None, type=None):
         filename = os.path.basename(source_key)
         filename = update_filename_extension(filename, type)
         destination_key = f"{self.upload_dir}/{prefix}/{filename}"
+        content_type = get_content_type(filename)
```
pamfilos: can we use the `type` kwarg to find the mimetype? would be cleaner

pamfilos: lets also setup an integration test with the minio S3 setup on docker-compose

pamfilos (changes requested):

dags/common/scoap3_s3.py
```diff
@@ -146,13 +165,14 @@ def download_and_upload_to_s3(self, url, prefix=None, headers=None, type=None):
...
             return
-
+        content_type = get_content_type(filename, response.headers)
         try:
             # Upload the file to S3
             self.client.put_object(
                 Body=response.content,
                 Bucket=self.bucket,
                 Key=destination_key,
+                ContentType=content_type,  # <-- Injetado no Upload!
```
pamfilos: remove comment :)

dags/common/scoap3_s3.py
pamfilos: any updates here?

pamfilos (changes requested):

